PR #17, opened 2021-09-24
added full bech32 capabilities + tests

nemo83:
# Bech32

In this PR I'm enhancing Bech32 capabilities so to be able to deal with all the Cardano addresses (stake, wallet, pool) directly in Java w/o needing the JNI bridge. 

The PR is inspired from the .NET library https://github.com/CardanoSharp/cardanosharp-wallet/blob/1e9d1b6b3489b94422a36b139600475ced400082/CardanoSharp.Wallet/Encoding/Bech32.cs

Tests are included too.

This is still a draft, as the new class needs to be better tested/polished and integrated w/ the codebase. I'm opening the PR to let reviewer start pointing major issues if any.

Thanks,
Gio

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,6 +27,7 @@ dependencies {
     annotationProcessor 'org.projectlombok:lombok:1.18.20'
 
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
+    testImplementation "org.junit.jupiter:junit-jupiter-params:5.6.0"
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
     testImplementation 'org.hamcrest:hamcrest-library:2.2'
     testImplementation 'org.mockito:mockito-inline:3.7.7'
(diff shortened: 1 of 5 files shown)


nemo83:
* I've removed the `var` keyword that didn't seem to be supported.
* Tests for asset id are not passing, the bech32 is significantly different, will have to investigate the difference.
* I've successfully replaced bech32 decoding in the code so to not make use of the rust lib (test passing) I'm unsure though how to replace encoding as I'm not sure where the get the hrp.

satran004:
Thanks Gio.

Yes, asset id test cases are failing. 

In previous version of Bech32,  asset id encoding was fine but decode() for address was failing. I think that was the reason for JNI call for address to bytes method. So need to find the root cause.

For encoding method (bytes to address), I think, hrp will be decided based on  header bytes. This CIP may help https://cips.cardano.org/cips/cip19/

matiwinnetou:
@nemo83 any chance to wrap this up...

satran004:
@nemo83 fyi, I have added Bech32.java in this PR to the master branch. This is used in the current address generation logic in Java. So far it seems ok. Though it needs more test scenarios.

https://github.com/bloxbean/cardano-client-lib/blob/master/src/main/java/com/bloxbean/cardano/client/crypto/Bech32.java

nemo83:
Thanks for this. I'm gonna add more tests in the upcoming days. Holidays have been crazy.

Thanks!

PR #19, opened 2021-10-05
Add compiled libraries for Android

andyundso:
With this PR, the Rust library that this project uses will be compiled for Android in x86 and ARM for both 32 bit and 64 bit. I'm working on an Android app with a friend and we need to sign transactions, therefore this library was ideal.

A few notes on the implementation:

* I use `cargo ndk` for the compilation against Android
* I only built the Rust libraries for Android, building a Gradle package didn't make sense since we cannot directly run the tests on Android
* I also simplified the `LibraryUtils` class to use JNA's Platform class. This is also due to Android having another mechanism to load libraries.

I wasn't able yet to test it in a bigger extend: I ran it in the Android Emulator so far and will try out on an ARM 32-bit device somewhere in the next week, but I cannot give feedback if both 64bit builds work. But I thought you might appreciate the contribution.


satran004:
Great. @andyundso Thanks a lot  for the contribution.

Android compatibility is one of the common request from other devs. The github action seems to be working fine for android specific libs. I am going to merge this pull request now.

Let me know if you face any issue during your testing. I am also not sure how to test on other Android platforms. If you have any pointer, feel free to drop a message.

PR #59, opened 2022-01-05
Payment Transaction Improvements

nemo83:
This PR aims to collate all payments towards the same address together.

1. created `plus` (happy to get renamed to `add` or whatever makes sense) to the `Asset`, `MultiAsset` and `Value` classes
2. created a new, `createReceiverOutputsAndPopulateCostV2` method, this should/wil manage all the tx together and `merge` these going to the same address.

This should save in fees thanks to the creation of a single Value from the multiple Value objects creaated from a PAymentTransaction object.

It should also allow to solve the problem of this tx: https://cardanoscan.io/transaction/61f7f2d030b90344f9e3e6a0fabdfbedb13abc94ba33bc3453e6604fcdfd7c7f

Where I want to send a specific amount of ada (greated than the minAda) along with a native asset. In the tx above, despite having sent enough ada to the recipient address, I've still paid the min ada to send along with the native asset.

I haven't tested the PR yet, so please bear with me :-D


satran004:
Thanks a lot @nemo83 . Great enhancement. Just had a quick look. I will go through the PR in detail tomorrow.
 
 "plus" sounds ok to me. 

Yes, let's add some tests. You may also want to add an integration test. 

Also, you may need to do a pull to update your branch. I just noticed "cardano-serialization-lib" sub module is still there in your fork. That dependency has already been removed.

nemo83:
Thanks for the quick response.

Gave that a go and I'm getting a value not conservative .

I think the problem is that fee at payment transaction object does not make
sense anymore now, but should be an independent parameter.

Will rebase from upstream

satran004:
Thanks @nemo83 

1. I just noticed that the fee is quite high for the transaction that you had shared earlier.  Can you please confirm if you are setting final fee only in one PaymentTransaction ? 

In case of multiple PaymentTransactions, you  need to set the fee  only in one transaction. Otherwise n * fee will be paid.

Here's a test case for your reference :
https://github.com/bloxbean/cardano-client-lib/blob/8284f9c7363b3494a3741cf7286c20e79b3a208d/src/integration-test/java/com/bloxbean/cardano/client/backend/api/helper/TransactionHelperServiceIT.java#L301

2. > I think the problem is that fee at payment transaction object does not make sense anymore now, but should be an independent parameter.

Agreed. We can plan to deprecate that and remove it later.

3. Also, I am  thinking if someone has added two separate PaymetTransactions to one output address, one with ADA and other one with MultiAsset, should we always combine it ? Or, should we keep it optional (Like user may want to keep it separate for some other reason..) ?

We can explore if somehow we can pass a flag "consolidateOutputs" or "combineOutputValue" ... and then call this optimization method if the flag is set.  One option, we can add this flag to "TransactionDetailsParams" 

Let me know your thoughts.

nemo83:
1. No 🤦 I set the fee to each PaymentTransaction
2. Awesome taht we are on the same page, I fear though the change might be quite extended. Maybe worth to issue a different PR for that.
3. While I do get your point, I'm not sure that _sending to different indexes of the same address_ makes sense. If you want to keep things separated, a user would send to two addresses of the same wallet. one with the ada alone and one with the asset + the min ada. Sending to same address in different utxos, as far as I can think, has no benefit, only extra costs. Am I missing something?

satran004:
> 2\. Awesome taht we are on the same page, I fear though the change might be quite extended. Maybe worth to issue a different PR for that.

Yes, a separate PR makes sense.



> 3\. Sending to same address in different utxos, as far as I can think, has no benefit, only extra costs. Am I missing something?

You are right. I was overthinking. If someone wants more control, then low level serialization api is there.  To keep it simple, high level api can behave in a certain way.

src/main/java/com/bloxbean/cardano/client/util/Tuple.java
```diff
@@ -1,11 +1,28 @@
 package com.bloxbean.cardano.client.util;
 
+import java.util.Objects;
```
nemo83, line 3: @satran004 what's the best way of fixing this? 

I need `equals` and `hashcode` for the grouping to work properly. Should we introdue lombok instead?

Or leave it as is and create a test with the description so that if they get deleted the test will remind why we need them?

nemo83:
@satran004 struggling to run tests on Mac Book Pro x86_64.

I've run the `. scripts/build-mac-x86_64.sh` script.

I get this error on:
```
java.lang.UnsatisfiedLinkError: Error looking up function 'signExtended': dlsym(0x214644300, signExtended): symbol not found
	at com.sun.jna.Function.<init>(Function.java:252)
	at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:600)
	at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:576)
	at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:562)
	```

**Comment by satran004** [comment]
> I've run the `. scripts/build-mac-x86_64.sh` script.
> 
> I get this error on:

@nemo83 I tried on mac x86  with your repo. It's working for me though 3 tests are failing but different error.

Can you please delete native, rust/target folders under root folder and try to build rust module again ?

```
rm -rf native/ rust/target/
```

File: src/main/java/com/bloxbean/cardano/client/util/Tuple.java (hunk 1 above)
Comment (satran004) [reply src/main/java/com/bloxbean/cardano/client/util/Tuple.java:3]: @nemo83 We can introduce Lombak if that solves your problem. We are using Lombok in other places.

**Comment by nemo83** [comment]
> @satran004 struggling to run tests on Mac Book Pro x86_64.
> 
> I've run the `. scripts/build-mac-x86_64.sh` script.
> 
> I get this error on:
> 
> ```
> java.lang.UnsatisfiedLinkError: Error looking up function 'signExtended': dlsym(0x214644300, signExtended): symbol not found
> 	at com.sun.jna.Function.<init>(Function.java:252)
> 	at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:600)
> 	at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:576)
> 	at com.sun.jna.NativeLibrary.getFunction(NativeLibrary.java:562)
> 	```
> 
> Any idea?
> ```

Thanks for the reply, you pointed me in the right direction, not a rust expert, but a `cargo update` fixed the problem for me.

After cleaning the folders you mentioned, `cargo update` and regenerated the lib, I was able to run again tests both from cli and IDE.

I also fixed all the tests, updating them in line with the rationale of the test, I've added comments in tests for clarity.

I hope I will find time for some additional testing and a couple of integration tests.

I'm overall happy with the outcome so far.

File: rust/Cargo.lock [hunk 2]
```diff
@@ -85,9 +88,9 @@ dependencies = [
 
 [[package]]
 name = "ed25519"
-version = "1.3.0"
+version = "1.2.0"
```
nemo83, line 91: @satran004 this is what was bombing out. I needed a cargo update to generate the lib

src/test/java/com/bloxbean/cardano/client/backend/api/helper/FeeCalculationServiceTest.java
```diff
@@ -114,7 +110,7 @@ void testCalculateFeeSimplePaymentTransaction() throws ApiException, IOException
     void testCalculateFeeMultiplePaymentTransactions() throws ApiException, IOException, AddressExcepion, CborSerializationException {
 
         String receiver1 = "addr_test1qqwpl7h3g84mhr36wpetk904p7fchx2vst0z696lxk8ujsjyruqwmlsm344gfux3nsj6njyzj3ppvrqtt36cp9xyydzqzumz82";
-        String receiver2 = "addr_test1qqwpl7h3g84mhr36wpetk904p7fchx2vst0z696lxk8ujsjyruqwmlsm344gfux3nsj6njyzj3ppvrqtt36cp9xyydzqzumz82";
+        String receiver2 = "addr_test1qpchc7f6zln5c2ruarssn9p8vulhhn3rk7cvz4lm05eq2xwx8e7tf5l25n4ek3gch8tnj4k5z4236fvy6yndrusf7x0sv287t5";
```
nemo83, line 113: @satran004 in order to fix the test maintaining the rationale, I've changed the receiver address to something different, this demonstrated the backward compatibility of my changes.

src/test/java/com/bloxbean/cardano/client/backend/api/helper/UtxoTransactionBuilderTest.java
```diff
@@ -346,7 +347,7 @@ public void testBuildTransactionWithMultiAssetAndNotSufficientLovelaceWillReturn
                 .sender(sender)
                 .fee(BigInteger.valueOf(12000))
                 .unit(LOVELACE)
-                .amount(BigInteger.valueOf(1700000))
+                .amount(BigInteger.valueOf(3000000))
```
nemo83, line 350: @satran004 as one less utxo was created by the transaction builder, there were about 1.3 additional $ada required to use an additional utxo. I've just increased the amount to spend of a comparable value, this had the desired effect of consuming 3 utxos, as per the original intent of the test.

src/test/java/com/bloxbean/cardano/client/backend/api/helper/UtxoTransactionBuilderTest.java
```diff
@@ -365,13 +366,13 @@ public void testBuildTransactionWithMultiAssetAndNotSufficientLovelaceWillReturn
         assertThat(transaction.getBody().getInputs().get(1).getTransactionId(), is("aaaaaa341088ca1c0ed2384a3139d34a1de4b31ef6c9cd3ac0c4eb55108fdf85"));
         assertThat(transaction.getBody().getInputs().get(2).getTransactionId(), is("bbbbbb341088ca1c0ed2384a3139d34a1de4b31ef6c9cd3ac0c4eb55108fdf85"));
 
-        assertThat(transaction.getBody().getOutputs(), hasSize(3));
+        assertThat(transaction.getBody().getOutputs(), hasSize(2));
```
nemo83, line 369: @satran004 here the result is well different, my changes reduce the number of created outputs, hence one less output. 

same as below change.

src/test/java/com/bloxbean/cardano/client/backend/api/helper/UtxoTransactionBuilderTest.java
```diff
@@ -677,7 +678,7 @@ public void testBuildTransaction_whenNotEnoughUtxosLeft() throws ApiException, I
         assertThat(transaction.getBody().getInputs().get(1).getTransactionId(), is("f3c464be15a5e29a1a6d322c5cd040c87075d1cfc89d4b397568d14c0ba53cd9"));
 
         assertThat(transaction.getBody().getOutputs().get(0).getValue().getCoin(), is(newAmount));
-        assertNull(transaction.getBody().getOutputs().get(0).getValue().getMultiAssets());
+        assertThat(transaction.getBody().getOutputs().get(0).getValue().getMultiAssets(), is(empty()));
```
nemo83, line 681: @satran004 I replaced nulls with empty lists.

src/test/java/com/bloxbean/cardano/client/backend/api/helper/FeeCalculationServiceTest.java
satran004: Thanks. Makes sense. Not sure why both were same before :)

src/test/java/com/bloxbean/cardano/client/backend/api/helper/UtxoTransactionBuilderTest.java
satran004: Cool. Hence less fee :)

satran004:
> I hope I will find time for some additional testing and a couple of integration tests.

@nemo83 Great progress. Let me know once it's ready for merge.

Fyi, To run integration tests in IDE, you can set the project id in this class BFBaseTest.java.
When you are running all integration tests, occasionally you may find some test case failure which is due to  network condition during that time. Just re-run the failed test case in that case.

nemo83:
> > I hope I will find time for some additional testing and a couple of integration tests.
> 
> @nemo83 Great progress. Let me know once it's ready for merge.
> 
> Fyi, To run integration tests in IDE, you can set the project id in this class BFBaseTest.java. When you are running all integration tests, occasionally you may find some test case failure which is due to network condition during that time. Just re-run the failed test case in that case.

I've added just one IT as unit tests do already a pretty good job for the type of change I did.

I think the PR is ready to be merged if no objections or issue in the review?

nemo83:
Also for some reasons the contract IT are failing on branch as well as aon master. so I geuss it's a problem with the tests?

src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBuilderImpl.java
```diff
@@ -378,6 +382,82 @@ private void checkAndAddAdditionalUtxosIfMinCostIsNotMet(Map<String, Set<Utxo>>
...
+
+                    // Calculating if extra costs are required (diff between minAda and actual ada, and add to costs)
+                    BigInteger additionalCost = actualCoin.subtract(draftTxOut.getValue().getCoin());
+                    existingMiscCost = existingMiscCost.add(additionalCost);
+
+                    BigInteger costs = entry.getValue().stream().map(PaymentTransaction::getFee).reduce(BigInteger.ZERO, BigInteger::add);
+                    existingMiscCost = existingMiscCost.add(costs);
+
+                    // Update costs per (sender) base address
+                    senderMiscCostMap.put(entry.getKey()._1.baseAddress(), existingMiscCost);
```
satran004, line 435: @nemo83, thanks for the optimization.

Are we missing  datumHash in case of transfer to a script address ? So if there is a datumHash  set in PaymentTransaction, it should be added to the output.

I am just thinking what would happen when there are two PaymentTransactions (to the same script address), with two different datum hash. Should we also group PaymentTransaction based on datumHash (sender, receiver, datumHash) ?  

Here's the link to existing code https://github.com/bloxbean/cardano-client-lib/blob/914ed1d16878cfa964b78783ddce4cf6b1e4ff2f/src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBuilderImpl.java#L415

We can merge this PR and create an issue to track as a separate PR.

Please let me know.

satran004:
> Also for some reasons the contract IT are failing on branch as well as aon master. so I geuss it's a problem with the tests?

Yes. There is an issue with collateral amount for one of the contract. I need to increase the amount. I will fix that after merge.

src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBuilderImpl.java
satran004: Or, ignore grouping in case there is datumhash. Probably a safer option.

satran004: @nemo83 Is it possible to ignore grouping when datumhash is there and just consider it as a separate output ?  In case of script output, I am not sure if we can add this intelligence. It depends on the  logic in validator script and output becomes non spendable based on the logic.

Probably safe to keep these output separate. 

BTW, I just tried contract IT tests. Those are failing because of missing datumHash. The contract call transaction doesn't find any utxo with datum hash as the previous payment transaction ignores datum hash in output :)

nemo83: Gotcha, it makes sense tests were failing. Of course the DatumHash wasnt set.

I'm gonna propose some changes. I believe grouping can still be safely done.

src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBuilderImpl.java
```diff
@@ -33,6 +37,14 @@
...
 public class UtxoTransactionBuilderImpl implements UtxoTransactionBuilder {
 
+    @Data
+    @AllArgsConstructor
+    private class PaymentTransactionGroupingKey {
+        private Account sender;
+        private String receiver;
+        private String datumHash;
+    }
+
```
nemo83, line 47: This `case class` identifies the transaction grouping rationale. Basically in order to collate transactions in just one input:
1. sender
2. receiver
3. datum (hash)
need to be the same.

src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBuilderImpl.java
nemo83: Right I've suggested some changed for the datum hash.

Pleasee check code [here](https://github.com/bloxbean/cardano-client-lib/pull/59/files#r781539951).

Although all the smart contracts tests are now passing, the following is constantly failing, in both master and this branch

`TransactionHelperServiceIT. mintTokenWithNormalAccountKeyHash()`

I will try to take a look at it tomorrow.

satran004: @nemo83 Please ignore the test failure for now. I think it is due to another check-in. I will fix it in the master after merge.

satran004:
@nemo83 Merging this PR. It looks good to me. Thanks a lot for this enhancement. Less fee is always better :)

nemo83:
Thanks for the hard work. It's only my pleasure being able to contribute to a good opensource project. Even more if it's for cardano!

PR #63, opened 2022-01-09
Policy as new Field instead of deprecated policyScript and policyKeys Fields in MintTransaction Object + PolicyUtil based on https://github.com/input-output-hk/cardano-node/blob/master/doc/reference/simple-scripts.md

edridudi:


src/main/java/com/bloxbean/cardano/client/util/PolicyUtil.java
```diff
@@ -0,0 +1,60 @@
...
+import com.bloxbean.cardano.client.transaction.spec.script.ScriptAll;
+import com.bloxbean.cardano.client.transaction.spec.script.ScriptAtLeast;
+import com.bloxbean.cardano.client.transaction.spec.script.ScriptPubkey;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PolicyUtil {
+
+    private static final int SLOTS_PER_EPOCH = 5 * 24 * 60 * 60;
```
satran004, line 19: @edridudi Should we get this from Genesis info instead of hard coded value ?

[NetworkInfoService.getNetworkInfo()](https://github.com/bloxbean/cardano-client-lib/blob/master/src/main/java/com/bloxbean/cardano/client/backend/api/NetworkInfoService.java#L13)

We can create a separate issue to handle this.

src/main/java/com/bloxbean/cardano/client/util/PolicyUtil.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public static Policy createEpochBasedTimeLockedPolicy(String name, long currentSlot, long epochs) throws CborSerializationException {
```
satran004, line 21: Very useful helper class. It will help to remove current boilerplate code.

src/main/java/com/bloxbean/cardano/client/util/PolicyUtil.java
edridudi: Good point. 
It's not crucial because SlotsPerEpoch is not about to change in the new future but yes I will create an issue for that.

PR #65, opened 2022-01-12
Random-Improve Coin Selection Strategy Enhancement

edridudi:


build.gradle
```diff
@@ -88,7 +88,7 @@ task fatJar(type: Jar) {
     manifest {
         attributes 'Main-Class': 'com.bloxbean.cardano.client.cli.CLIMain'
     }
-    baseName = project.name + '-all'
+    archivesBaseName = project.name + '-all'
```
satran004, line 91: @edridudi Can we ignore these changes as part of this PR? Too many files. Hard to review :)  Let's keep the change to files specific to this PR.

scripts/build-linux-x86_64.sh
```diff
@@ -2,7 +2,7 @@ export SRC_LIB_FILE=libcardano_jni_wrapper.so
 export TARGET_LIB_FILE=linux-x86-64_libcardano_jni_wrapper.so
 export NATIVE_FOLDER=linux-x86-64
 
-cd rust
+cd ../rust
```
satran004, line 5: @edridudi This will break the step in README and CI.  build-linux-*.sh is invoked from root folder.

build.gradle
edridudi: of course ..

scripts/build-linux-x86_64.sh
edridudi: Oh I see.. It didn't work for me when I tried running it locally hence the change.

satran004: Thanks. No worries. It would be good if you can just rollback files which are not specific to this PR. I will then merge this PR to the branch.

PR #68, opened 2022-01-18
Allow to create an Account from a private key

bcoste:
Allow to create an Account from a private key because it's not always possible to have the seed available and it's safer/faster to derive from the account hardened key.

### DIFF
--- a/src/main/java/com/bloxbean/cardano/client/account/Account.java
+++ b/src/main/java/com/bloxbean/cardano/client/account/Account.java
@@ -27,7 +27,10 @@ import java.util.stream.Collectors;
 public class Account {
     @JsonIgnore
     private String mnemonic;
+    @JsonIgnore
+    private String secretKey;
     private String baseAddress;
+    private String changeAddress;
     private String enterpriseAddress;
     private String stakeAddress;
     private Network network;
@@ -131,8 +134,27 @@ public class Account {
     public Account(Network network, String mnemonic, DerivationPath derivationPath) {
         this.network = network;
         this.mnemonic = mnemonic;
+        this.secretKey = null;
         this.derivationPath = derivationPath;
         validateMnemonic();
+        baseAddress();
+    }
+
+    /**
+     * Create an account for the network and derivation path from a private key
+     *
+     * @param network
+     * @param derivationPath
+     * @param secretKey is a private key of 192 bytes or 256 bytes (with pubkey and chaincode) encoded in a base64 String
+     */
+    public Account(Network network, DerivationPath derivationPath, String secretKey) {
+        this.network = network;
+        this.mnemonic = null;
+        if (secretKey.length() == 192)
+            this.secretKey = secretKey;
+        else
+            this.secretKey = secretKey.substring(0,128) + secretKey.substring(192,256);
+        this.derivationPath = derivationPath;
         baseAddress();
     }
 
@@ -148,7 +170,7 @@ public class Account {
      */
     public String baseAddress() {
         if (baseAddress == null || baseAddress.isEmpty()) {
-            HdKeyPair paymentKeyPair = getHdKeyPair();
+            HdKeyPair paymentKeyPair = getExternalKeyPair();
             HdKeyPair stakeKeyPair = getStakeKeyPair();
 
             Address address = AddressService.getInstance().getBaseAddress(paymentKeyPair.getPublicKey(), stakeKeyPair.getPublicKey(), network);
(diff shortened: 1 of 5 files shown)


bcoste:
For the failing tests :
- testGetBaseAddressFromMnemonicByNetworkInfoMainnet 
- testGetBaseAddressFromMnemonicByNetworkInfoTestnet
- testGetBaseAddressFromMnemonicIndex_1
- testGetBaseAddressFromMnemonicIndex_2

This is due to my changes in the baseAddress, I force to calculate it from the external address of index 0.

From my understanding, base address represents the base index 0 of address_index in a BIP44 (and also CIP 1852) path :
m / purpose' / coin_type' / account' / change / address_index

If I'm right, the tests try to calculate a base index with a value greater than 0 instead of 0, and it should not be possible.

Are you ok to change the baseAddress constant value and index to 0 ?

Is it possible to introduce a currentAddress property to represent the address matching the index in the derivation path of the Account ?

Best regards,

Bruno

satran004:
@bcoste Thanks a lot for the PR. It's a good enhancement.

>> This is due to my changes in the baseAddress, I force to calculate it from the external address of index 0.

I am wondering if we really need to force it to index 0. I noticed you are forcing it to 0 in two places,  during baseAddress generation and sign().  

If there is a utxo from a different address (let's say account=0, index=1) which is part of the input, I am not sure if the sign() will work. Because, you may need to sign the transaction with a private key from the address at index=1 to spend the fund. That's my understanding. 


>>From my understanding, base address represents the base index 0 of address_index in a BIP44 (and also CIP 1852) path :
m / purpose' / coin_type' / account' / change / address_index


As per my understanding, baseAddress where header_type is "0000" contains both paymentKey and StakeKey. But it doesn't talk about derivation path.  (plz check this cip: https://cips.cardano.org/cips/cip19/)

The baseAddress section in this doc also doesn't mention about cip1852 index = 0. https://docs.cardano.org/core-concepts/cardano-addresses

Same with cardano-serialization-lib rust impl https://github.com/Emurgo/cardano-serialization-lib/blob/6a8c67a0fdb2ac6d72132d9693b87a6734f3fed5/rust/src/address.rs#L488

>> Is it possible to introduce a currentAddress property to represent the address matching the index in the derivation path of the Account ?

I am not sure if it's good idea to introduce another address type "currentAddress". But if my understanding of baseAddress is not correct, we can definitely introduce that.


I just tried changing getExternalKeyPair() method to getHdKeyPair() in both Account.baseAddress(), Account.sign(). All tests are working fine after that change. I also tried a txn with utxos from index=2 and txn signing with HdPrivateKey at index=2.

Am I missing something?

src/main/java/com/bloxbean/cardano/client/account/Account.java
```diff
@@ -278,13 +315,50 @@ private void validateMnemonic() {
...
+        if (mnemonic == null || mnemonic.trim().length() == 0) {
+            hdKeyPair = new CIP1852().getKeyPairFromAccountKey(this.secretKey, derivationPath);
+        }
+        else {
+            hdKeyPair = new CIP1852().getKeyPairFromMnemonic(mnemonic, derivationPath);
+        }
+        return hdKeyPair;
+    }
+
+    public HdKeyPair getExternalKeyPair() {
```
satran004, line 328: Also,  we can change this method to private as it's not used  outside of this class. 
Currently if you print "Account" object, it shows privatekey and publickey in the log which may not be desirable.  If we really need to make this method public, we can add @JsonIgnore annotation to this method.

satran004:
or, should we keep it simple for this PR. so for secret key, it is always account=0, index=0.

So, only one constructor in Account class for secret key. I don't know if we go with the current constructor with derivation path, if there will be any issue with changeAddress. May be, for this PR, we can keep it to index=0 for secret key and then extend it later as we learn more.

```
 public Account(Network network, String secretKey)
```

src/main/java/com/bloxbean/cardano/client/account/Account.java
bcoste: I left this method public because I wanted to extract the public key to sign payloads of a transaction to give it to constructionCombine of cardano-rosetta.
But you're right, it's better to keep it private to have a good implementation.

bcoste:
Thanks for your fast answer !
After checking the cardano docs and rust code, I agree with you on the fact a baseAddress can have an address-index index different from 0.
The changes you made are also fine from my point of view.
If we can just remove DerivationPath.createAccountDerivationPath() because I think it's not useful and may give wrong results ?

satran004:
@bcoste Good to know that we are in sync now :)

It would be great if you can update the PR as discussed.  I will merge it after that.

- Remove derivation path from the new constructor for secret key in Account class
- Use getHdKey() in baseAddress() and sign() in Account class
- Remove getExternalHDKey() if not required.
- Remove any other methods which are not required. (Exp: DerivationPath.createAccountDerivationPath())


Thanks again for your contribution.

satran004 (approved):
Thanks @bcoste. Looks good to me. I will do some minor refactoring after merge. Like account secret key as byte[] in constructor instead of String.

PR #80, opened 2022-02-22
Refactored UtxoSelectionStrategy + added RandomImproveUtxoSelectionStrategy

stikkos:
- Refactored UtxoSelectionStrategy interface to accept multiple assets and max utxo selection limit which is required for random - improve algorithm
- Added RandomImproveUtxoSelectionStrategy (including test) with fallback to LargestFirstUtxoSelectionStrategy
- Added LargestFirstUtxoSelectionStrategy (including test)
- Added UtxoTransactionBodyBuilder to work with new UtxoSelectionStrategy

### DIFF
--- a/src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBodyBuilder.java
+++ b/src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBodyBuilder.java
@@ -1,6 +1,5 @@
 package com.bloxbean.cardano.client.backend.api.helper.impl;
 
-import com.bloxbean.cardano.client.address.Address;
 import com.bloxbean.cardano.client.backend.exception.InsufficientBalanceException;
 import com.bloxbean.cardano.client.backend.model.Amount;
 import com.bloxbean.cardano.client.backend.model.ProtocolParams;
@@ -8,16 +7,12 @@ import com.bloxbean.cardano.client.backend.model.Utxo;
 import com.bloxbean.cardano.client.coinselection.UtxoSelectionStrategy;
 import com.bloxbean.cardano.client.common.CardanoConstants;
 import com.bloxbean.cardano.client.common.MinAdaCalculator;
-import com.bloxbean.cardano.client.common.model.Networks;
 import com.bloxbean.cardano.client.transaction.model.MintTransaction;
 import com.bloxbean.cardano.client.transaction.model.PaymentTransaction;
 import com.bloxbean.cardano.client.transaction.model.TransactionDetailsParams;
 import com.bloxbean.cardano.client.transaction.model.TransactionRequest;
 import com.bloxbean.cardano.client.transaction.spec.*;
-import com.bloxbean.cardano.client.util.AssetUtil;
-import com.bloxbean.cardano.client.util.HexUtil;
-import com.bloxbean.cardano.client.util.Triple;
-import com.bloxbean.cardano.client.util.Tuple;
+import com.bloxbean.cardano.client.util.*;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
@@ -34,23 +29,12 @@ public class UtxoTransactionBodyBuilder {
     public static TransactionBody buildMintBody(MintTransaction request,
                                                 TransactionDetailsParams detailsParams,
                                                 ProtocolParams protocolParams,
-                                                UtxoSelectionStrategy selectionStrategy,
-                                                NetworkId networkType) {
+                                                UtxoSelectionStrategy selectionStrategy) {
         String sender = request.getSender().baseAddress();
 
         String receiver = request.getReceiver();
         if (receiver == null){
             receiver = sender;
-        }
-        {
-            var senderNetwork = determineNetwork(sender);
-            var receiverNetwork = determineNetwork(receiver);
-            if(!networkType.equals(senderNetwork)){
-                throw new IllegalArgumentException("Network type for [" + sender + "] invalid. Expected [" + networkType + "] but was [" + senderNetwork + "]");
-            }
-            if(!networkType.equals(receiverNetwork)){
-                throw new IllegalArgumentException("Network type for [" + receiver + "] invalid. Expected [" + networkType + "] but was [" + receiverNetwork + "]");
-            }
         }
 
         var minBaseAmount = new MinAdaCalculator(protocolParams)
@@ -101,27 +85,10 @@ public class UtxoTransactionBodyBuilder {
                 null);  // TODO add network?
     }
 
-    public static NetworkId determineNetwork(String address){
-        return Networks.mainnet().equals(new Address(address).getNetwork()) ? NetworkId.MAINNET : NetworkId.TESTNET;
-    }
-
     public static TransactionBody buildTransferBody(List<PaymentTransaction> requests,
                                                     TransactionDetailsParams detailParams,
                                                     ProtocolParams protocolParams,
-                                                    UtxoSelectionStrategy selectionStrategy,
-                                                    NetworkId networkType){
-        // verify senders and receivers
-        for(var request : requests){
-            var senderNetwork = determineNetwork(request.getSender().baseAddress());
-            var receiverNetwork = determineNetwork(request.getReceiver());
-            if(!networkType.equals(senderNetwork)){
-                throw new IllegalArgumentException("Network type for [" + request.getSender().baseAddress() + "] invalid. Expected [" + networkType + "] but was [" + senderNetwork + "]");
-            }
-            if(!networkType.equals(receiverNetwork)){
-                throw new IllegalArgumentException("Network type for [" + request.getReceiver() + "] invalid. Expected [" + networkType + "] but was [" + receiverNetwork + "]");
-            }
-        }
-
+                                                    UtxoSelectionStrategy selectionStrategy){
         // Create output for receivers and calculate total fees/cost for each sender
         var calculatedOutputs = calculateOutputs(requests, protocolParams);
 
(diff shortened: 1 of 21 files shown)


src/main/java/com/bloxbean/cardano/client/coinselection/UtxoSelectionStrategy.java
```diff
@@ -12,42 +15,27 @@
...
-    /**
-     * Selected utxos based on a strategy
-     * @param address Address to select utxos for
-     * @param unit Unit
-     * @param amount Amount
-     * @param excludeUtxos Utxos to ignore
-     * @return List of Utxos
-     */
-    List<Utxo> selectUtxos(String address, String unit, BigInteger amount, Set<Utxo> excludeUtxos) throws ApiException;
+    int DEFAULT_COIN_SELECTION_LIMIT = 20;
```
satran004, line 18: We can keep DEFAULT_COIN_SELECTION_LIMIT constant in Configuration class with a setter in case application wants to set to higher or lower value. What do you think ?

Similarly, the default selection strategy can be kept in Configuration. So the default strategy can be changed through Configuration by the application. Currently, the default strategy is created inside UtxoTransactionBuilderImpl's constructor.

src/main/java/com/bloxbean/cardano/client/coinselection/impl/DefaultUtxoSelectionStrategyImpl.java
```diff
@@ -1,116 +1,149 @@
...
+                var fetched = fetchResult != null && fetchResult.getValue() != null
+                        ? fetchResult.getValue()
+                                     .stream()
+                                     .sorted(sortByMostMatchingAssets(outputAmounts))
+                                     .collect(Collectors.toList())
+                        : Collections.<Utxo>emptyList();
+
+                page += 1;
+                for(Utxo utxo : fetched) {
+                    if(!accept(utxo)){
```
satran004, line 70: What's the use of this condition? It always returns true;

src/main/java/com/bloxbean/cardano/client/coinselection/impl/DefaultUtxoSelectionStrategyImpl.java
```diff
@@ -1,116 +1,149 @@
...
+            int page = 0;
+            final int nrOfItems = 100;
+
+            while(!remaining.isEmpty()){
+                var fetchResult = utxoService.getUtxos(sender, nrOfItems, page + 1, OrderEnum.asc);
+
+                var fetched = fetchResult != null && fetchResult.getValue() != null
+                        ? fetchResult.getValue()
+                                     .stream()
+                                     .sorted(sortByMostMatchingAssets(outputAmounts))
```
satran004, line 64: Should we start calling this default impl as "MostMatchedAssetSequentialStrategy" :)

satran004 (commented):
Looks good to me. Only few comments.

src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBodyBuilder.java
```diff
@@ -0,0 +1,549 @@
...
+            receiver = sender;
+        }
+
+        var minBaseAmount = new MinAdaCalculator(protocolParams)
+                .calculateMinAda(request.getMintAssets());
+        BigInteger totalCost = minBaseAmount.add(request.getFee());
+        // determine UTXOs for min cost
+        var utxos = request.getUtxosToInclude() != null ? new HashSet<>(request.getUtxosToInclude()) : new HashSet<Utxo>();
+        if(utxos.isEmpty()){
+            utxos.addAll(selectionStrategy.select(sender, new Amount(CardanoConstants.LOVELACE, totalCost), null, Collections.emptySet()));
```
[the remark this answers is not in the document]
stikkos: Yes, I tried to keep this as before. So all current selection strategy implementations should throw InsufficientBalanceException if there isn't enough funds.

src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBodyBuilder.java
```diff
@@ -0,0 +1,549 @@
...
+                transactionOutputs,
+                request.getFee(),
+                detailsParams != null ? detailsParams.getTtl() : 0L,
+                null,
+                detailsParams != null ? detailsParams.getValidityStartInterval() : 0L,
+                request.getMintAssets(),
+                null,
+                null,
+                null,
+                null);  // TODO add network?
```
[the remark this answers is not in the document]
stikkos: Yes, good idea, have added it in latest commit

src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBodyBuilder.java
```diff
@@ -0,0 +1,549 @@
...
+
+    // initialize base + multi assets with negative amount (based on what was requested)
+    private static OutputAmount initChangeAmount(List<Amount> requestedAmounts){
+        BigInteger baseAmount = BigInteger.ZERO;
+        List<MultiAsset> multiAssets = new ArrayList<>();
+        for(var amount : requestedAmounts){
+            if(CardanoConstants.LOVELACE.equals(amount.getUnit())) {
+                baseAmount = baseAmount.subtract(amount.getQuantity());
+            } else {
+                multiAssets.add(AssetUtil.getMultiAssetFromUnitAndAmount(amount.getUnit(), BigInteger.ZERO.subtract(amount.getQuantity())));
```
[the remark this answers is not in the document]
stikkos: Yes, well spotted. Have corrected this in latest commit

src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBodyBuilder.java
```diff
@@ -0,0 +1,549 @@
...
+                    || !changeAmount.getMultiAssets().isEmpty()) {
+                BigInteger minRequiredLovelaceInOutput =
+                        new MinAdaCalculator(protocolParams)
+                                .calculateMinAda(new TransactionOutput(sender, new Value(changeAmount.getBaseAmount(), changeAmount.getMultiAssets())));
+                // then select UTXO we can use for this extra output
+                while(minRequiredLovelaceInOutput.compareTo(changeAmount.getBaseAmount()) > 0) {
+                    //Get utxos
+                    var allUsedUtxos = new HashSet<>(usedUtxos);
+                    allUsedUtxos.addAll(additionalUtxos);
+                    Set<Utxo> newUtxos = calculateUtxos(sender, Collections.singletonList(new Amount(CardanoConstants.LOVELACE, minRequiredLovelaceInOutput)), allUsedUtxos, selectionStrategy);
```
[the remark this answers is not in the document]
stikkos: Yes, not sure if it matters but I think you are right. Have corrected it in latest commit.

src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/UtxoTransactionBuilderImpl.java
```diff
@@ -160,525 +90,29 @@ public Transaction buildTransaction(List<PaymentTransaction> transactions, Trans
...
-
-    private BigInteger createDummyOutputAndCalculateMinAdaForTxnOutput(String address, List<MultiAsset> multiAssets, ProtocolParams protocolParams) {
-        TransactionOutput txnOutput = new TransactionOutput();
-        //Dummy address
-        txnOutput.setAddress(address);
-        txnOutput.setValue(new Value(BigInteger.ZERO, multiAssets));
-
-        return new MinAdaCalculator(protocolParams).calculateMinAda(txnOutput);
+    @Override
+    public List<Utxo> getUtxos(String address, String unit, BigInteger amount) throws ApiException {
```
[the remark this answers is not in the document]
stikkos: added `Deprecated` to interface

src/main/java/com/bloxbean/cardano/client/coinselection/UtxoSelectionStrategy.java
stikkos: Have moved `DEFAULT_COIN_SELECTION_LIMIT` to `Configuration` but `DefaultUtxoSelectionStrategyImpl` requires a utxoSevice to construct.

src/main/java/com/bloxbean/cardano/client/coinselection/impl/DefaultUtxoSelectionStrategyImpl.java
stikkos: :-) Yes, just kept this, might be useful to override a strategy and filter out some UTXOs?

src/main/java/com/bloxbean/cardano/client/coinselection/impl/DefaultUtxoSelectionStrategyImpl.java
stikkos: The reason I added `sort` was to mimic the existing logic. 

In the current `UtxoTransactionBuilderImpl` on master, when processing a multi-asset `PaymentTransaction`, first the utxos are selected for the native token and only after that for min lovelace.

But for this PR, all assets are send to the selection strategy together and utxos are only fetched once.
But processing them in natural order could result in more tx's being selected. For example, if there is a UTXO with enough native tokens and ada to cover all requested amounts but another UTXO exist (before this one) with only ADA, then that one will be picked for covering the min ADA and the other one for covering the native token.

in order to achieve the same result as before, this sort was added, but I'll leave it up to you to rename (or change) as you wish :-)

satran004:
Merged. Thanks a lot @stikkos  !!!

Issue #79, opened 2022-02-20
Check if minimum Ada is not met - issue


sourcelib-stack:
Hi,

In the _**UtxoTransactionBuilderImpl**.buildOuputsForSenderFromUtxos_ you have this check
`
//Check if minimum Ada is not met. Topup
//Transaction will fail if minimum ada not there. So try to get some additional utxos
verifyMinAdaInOutputAndUpdateIfRequired(transactionInputs, changeOutput, detailsParams, utxoSet, protocolParams);
`

this is not quite correct as the rule is : transaction should be min 1 ada and not the resulting UTOX in the wallet.

For example if I have 1.5 ada in a wallet and I want to transfer it this should be fine (1.5 - fee). the node will validate it.

satran004:
Thanks.
>> this is not quite correct as the rule is : transaction should be min 1 ada and not the resulting UTOX in the wallet.

As per my understanding, resulting UTXO should have some min ada. Without multi-asset it is 1 ada.
Please check here https://docs.cardano.org/native-tokens/minimum-ada-value-requirement

>> For example if I have 1.5 ada in a wallet and I want to transfer it this should be fine (1.5 - fee). the node will validate it.
Yes node should validate that. In this case most probably there will be only 1 Utxo.

sourcelib-stack:
Just to clarify a bit, the issue I'm signaling here is that with this condition there you cannot send all the ada from a wallet, it throws the error.

If I comment out the verifyMinAdaInOutputAndUpdateIfRequired works just fine.

satran004:
Closing this. @sourcelib-stack  please re-open if you still see this issue.

PR #84, opened 2022-03-07
Fix [#79] - Retry fee calculation (with amount = amount - fee) in cas…

satran004:
To avoid exception during fee calculation -

- If fee calculation throws InsufficientBalanceException, retry fee calculation with an updated amount (amount = amount - fee) in catch block. 
- If retry still fails, then throw exception.

To send all ada from an address after successful fee calculation, the application has to update the lovelace amount 

amount =  current ada balance - calculated fee

**Example**  To send all ada from an address with balance 1.5 Ada

```
//Create a payment transaction with total lovelace balance
            PaymentTransaction paymentTransaction =
                    PaymentTransaction.builder()
                            .sender(receiver)
                            .receiver(sender.baseAddress())
                            .amount(adaToLovelace(1.5))
                            .unit(LOVELACE)
                            .build();

//Calculate fee
            BigInteger fee = feeCalculationService.calculateFee(paymentTransaction, TransactionDetailsParams.builder().ttl(getTtl()).build(), null);

            //Update the amount and fee
            paymentTransaction.setAmount(paymentTransaction.getAmount().subtract(fee));
            paymentTransaction.setFee(fee);

//Submit transaction
            result = transactionHelperService.transfer(paymentTransaction, TransactionDetailsParams.builder().ttl(getTtl()).build());
```

### DIFF
--- a/src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/FeeCalculationServiceImpl.java
+++ b/src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/FeeCalculationServiceImpl.java
@@ -58,16 +58,19 @@ public class FeeCalculationServiceImpl implements FeeCalculationService {
 
         PaymentTransaction clonePaymentTransaction = paymentTransaction.toBuilder().build();
         if(clonePaymentTransaction.getFee() == null || clonePaymentTransaction.getFee().compareTo(DUMMY_FEE) == -1) //Just a dummy fee
-            clonePaymentTransaction.setFee(DUMMY_FEE); //Set a min fee just for calcuation purpose if not set
+            clonePaymentTransaction.setFee(DUMMY_FEE); //Set a min fee just for calculation purpose if not set
 
         String txnCBORHash;
         try {
             //Build transaction
             txnCBORHash = transactionHelperService.createSignedTransaction(Arrays.asList(clonePaymentTransaction), detailsParams, metadata);
         } catch (InsufficientBalanceException e) {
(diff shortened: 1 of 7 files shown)


src/main/java/com/bloxbean/cardano/client/backend/api/helper/impl/FeeCalculationServiceImpl.java
```diff
@@ -47,10 +56,19 @@ public BigInteger calculateFee(PaymentTransaction paymentTransaction, Transactio
...
+        if(clonePaymentTransaction.getFee() == null || clonePaymentTransaction.getFee().compareTo(DUMMY_FEE) == -1) //Just a dummy fee
+            clonePaymentTransaction.setFee(DUMMY_FEE); //Set a min fee just for calcuation purpose if not set
+
+        String txnCBORHash;
+        try {
+            //Build transaction
+            txnCBORHash = transactionHelperService.createSignedTransaction(Arrays.asList(clonePaymentTransaction), detailsParams, metadata);
+        } catch (InsufficientBalanceException e) {
+            clonePaymentTransaction.setFee(MIN_DUMMY_FEE);
+            clonePaymentTransaction.setAmount(clonePaymentTransaction.getAmount().subtract(MIN_DUMMY_FEE));
```
stikkos, line 69: Is there no check required to make sure tx amount is lovelace?

satran004: Right.  I have added a condition to check if unit = lovelace then retry. If not lovelace, then throw original exception.
Thanks.

PR #87, opened 2022-03-09
Add Koios Backend Support

edridudi:


edridudi:
@satran004, Most Endpoints are supported. There are still some gaps between the APIs so created the PR to discuss them.
Will publish the list in a few hours.

* I'm Not sure why build failed

backends/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosAddressService.java
```diff
@@ -0,0 +1,122 @@
...
+            throw new ApiException("Failed to Parse Date: " + e.getMessage(), e);
+        }
+    }
+
+    private Result<List<AddressTransactionContent>> convertToAddressTransactionContent(List<TxHash> transactions) throws ParseException {
+        List<AddressTransactionContent> addressTransactionContents = new ArrayList<>();
+        for (TxHash txHash : transactions) {
+            AddressTransactionContent addressTransactionContent = new AddressTransactionContent();
+            addressTransactionContent.setTxHash(txHash.getTxHash());
+//            addressTransactionContent.setTxIndex(); TODO
```
edridudi, line 115: Won't be supported by Koios. Can be Fetched using another API call.

satran004:
Thanks. Merged. For non-supported attributes, let's add some comments later.

Issue #90, opened 2022-03-12
Move Koios Urls to a Constant class in koios main module


satran004:
Currently, the Constant.java is under "it" module. This class has all Koios related urls.

Should we move this class to koios main module if these urls are fixed ?

(Similar to com.bloxbean.cardano.client.backend.impl.blockfrost.common.Constants class for Blockfrost backend)

Issue #89, opened 2022-03-12
Koios backend failed integration test


satran004:
The following test is failing 
https://github.com/bloxbean/cardano-client-lib/blob/e7ca0f2ca3628797569b14a786107f24bc3647aa/backends/koios/src/it/java/com/bloxbean/cardano/client/backend/koios/it/KoiosMetadataServiceIT.java#L31

PR #91, opened 2022-03-12
Fix: Koios backend failed integration test (#89) +  Move Koios Urls to a Constant class in koios main module bloxbean (#90)

edridudi:


edridudi:
> @edridudi I mean "koios" subproject's main folder. Not root project's main folder. The above Constants class is for Blockfrost. So we need an equivalent one for Koios in Koios subproject.

what package?
com.bloxbean.cardano.client.backend.impl.koios.common ?

or should I create it in

com.bloxbean.cardano.client.backend.koios ?

satran004:
Thanks @edridudi. Looks good to me

PR #93, opened 2022-03-18
Updated project structure to avoid core module's direct dependency with Backend service apis

satran004:
### Why ?

Currently, cardano-client-lib main module has both core classes (cbor, transaction building logic, high level api, functional api) and backend service (backend api + Blockfrost impl) together.  The other backend modules can be optionally added to the application if required.  

While this structure is ok, it also creates a tight coupling with backend service impl.  For example, if an application wants to avoid repetitive utxo fetching logic using a cache, it's hard to do that without introducing your own backend layer.

 Also,  not all backend apis are required for transaction building or fee calculation. For example,  though AssetService and AddressService  provide asset info or address info, those are not mandatory to build or submit transaction but those are nice to have apis. In the current implementation, it is also hard to identify mandatory backend apis.

So in the new project structure, there are three main modules. Each module can be separately added to application's pom.xml or build.gradle as a dependency.

### 1.  cardano-client-lib 

This is the main module which has all the core logic ( Cbor serialization, Low level api, High Level api, Functional api ) required to build transaction.

The following 3 interfaces are introduced to avoid any direct dependency with backend apis. These interfaces provide an additional abstraction layer between core logic and backend service.

**- UtxoSupplier :**  Provides utxos at an address which are required to build a transaction. 

**- ProtocolParamsSupplier:** Provides network's current protocol parameters which are required during transaction building. (Min Ada calculation, Fee etc.)

**- TransactionProcessor:**  This interface has only one method which can be implemented to submit transaction to Cardano network.

To build / submit transactions, we need implementations of these interfaces. Applications can use out-of-box implementations of these interfaces which depend on backend service or can provide their own implementations.

### 2. Backend

 This module contains backend service api and Blockfrost backend implementation.

### 3. Backend modules:  

Backend implementations for different backends. Each backend has their own backend impl. (GraphQL, Koios etc.)


edridudi (approved):
Looks good!

PR #98, opened 2022-03-29
add function to get tx_hash from txBody

somereason:
add function to get tx_hash from txBody

### DIFF
--- a/src/main/java/com/bloxbean/cardano/client/transaction/spec/TransactionBody.java
+++ b/src/main/java/com/bloxbean/cardano/client/transaction/spec/TransactionBody.java
@@ -5,10 +5,13 @@ import com.bloxbean.cardano.client.exception.AddressExcepion;
 import com.bloxbean.cardano.client.exception.CborDeserializationException;
 import com.bloxbean.cardano.client.exception.CborSerializationException;
 import com.bloxbean.cardano.client.transaction.spec.cert.Certificate;
+import com.bloxbean.cardano.client.transaction.util.CborSerializationUtil;
 import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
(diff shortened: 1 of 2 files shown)


edridudi (approved):

src/main/java/com/bloxbean/cardano/client/transaction/spec/TransactionBody.java
```diff
@@ -53,18 +56,29 @@
 
     private NetworkId networkId; // 1 or 0
 
+    public String getHash() throws CborSerializationException {
```
satran004, line 59: Thanks @somereason !!!

Right now, auxiliaryDataHash field in txnBody is set inside Transaction.serialize(). So if someone tries to get transaction hash directly from TransactionBody before serialization of Transaction object, auxiliaryDataHash will be set to null and incorrect hash.

In your test it's ok as you are doing transaction sign and then creating txn hash.

Probably we have two options here to avoid this :

1. Move getHash() to Transaction class and before calculating hash, we can serialize and de-serialize the whole transaction. An extra serialization step, but generated hash would be the correct one.

2. Move getHash() to TransactionUtil.java.  The method can take a byte[] for serialized transaction.

My preference is option 2. But actually both are ok. What do u think ?

Finally, you can avoid using bouncycastle packages directly here. There is a helper method in KeyGenUtil which can be used to generate blake2b hash.

Please check the similar usage here 

https://github.com/bloxbean/cardano-client-lib/blob/9883c38ba35c970e9e5e85f42064f15e22709999/backend-modules/ogmios/src/main/java/com/bloxbean/cardano/client/backend/ogmios/OgmiosTransactionService.java#L69

somereason: I love option 2. I'll modify and pull again. 

The very reason my case work is ```account.sign(tx)``` will call ```serialize()```. It's pure lucky. 2333333333

PR #111, opened 2022-05-02
CIP22 Implementetion - Stake Pool Operator verification

nemo83:
Hello,

this is a draft PR for CIP22 for java.

While the code is pretty much straightforward, it does depend on two more moving parts:
1. A custom Cardano version of the libsodium library https://github.com/input-output-hk/libsodium
2. A customised version of the libsodium java library, drafted by Andrew Westberg https://github.com/AndrewWestberg/libsodium-jna/tree/iohk_version

The binary of the libsodium library can be easily compiled for multiple OS and architectures and attached to the project, but I'm not sure how to deal w/ the amended libsodium library. 

Recommendations/suggestions appreciated.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,6 +6,7 @@ plugins {
 
 repositories {
     mavenCentral()
+    mavenLocal()
 }
 
 dependencies {
@@ -18,6 +19,8 @@ dependencies {
 
     //ed25519
     implementation group: 'net.i2p.crypto', name: 'eddsa', version: '0.3.0'
+
+    compile 'com.muquit.libsodiumjna:libsodium-jna:1.1.0-IOG-SNAPSHOT'
 
     compileOnly 'org.projectlombok:lombok:1.18.20'
     annotationProcessor 'org.projectlombok:lombok:1.18.20'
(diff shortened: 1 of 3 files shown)


satran004:
Thanks @nemo83 for the PR. I will merge it to a separate branch for now. 
I would prefer to avoid any native lib dependency at this point. We can plan to merge it to main branch after 0.2.0 main release.

Few things to explore before that.
1. If any VRF Java impl is available, let's explore to use that. (Ideal solution)
2. Otherwise, check for any VRF rust impl. This cip can be part of an another extension module with native lib dependency (distributed as a separate Jar) in cardano-client project.
3. If no rust lib found, then use libsodium dependency but in an another extension module.

We can update the implementation in the branch based on the findings  and merge to main later.

nemo83:
Thanks for the feedback, I am in no pressure to get this merged, I just had the code lying around for a while and I made sense to make this PR.

[cncli](https://github.com/AndrewWestberg/cncli) is written in rust and implements the sign side of the cip22.

matiwinnetou:
It is closed because it is merged or it is invalid?

satran004:
The main reason was dependency with native library. We need to work on it later to find a pure java solution. If that's not possible, it can be implemented as a separate project.

matiwinnetou:
Got it, thanks for explanation.

PR #112, opened 2022-05-04
Remove JsonIgnore Annotation from getPolicyId()

edridudi:
Policy Id shouldn't be ignored because it is better not to add serialization capabilities in frontend apps (better performance)


satran004:
Thanks @edridudi

PR #113, opened 2022-05-05
Fix Koios Schema Change for Policy Assets Endpoint + Added Account Endpoint as Backend Service for Koios and BF

edridudi:


satran004:
Looks good. Thanks @edridudi

PR #116, opened 2022-05-08
Helper methods in DataSignature

satran004:
- Added methods to get address bytes, public key, curve from DataSignature
- Moved CIP30 values to a separate class
- Helper methods to get headers in HeaderMap


edridudi (approved):
Looks good.

Issue #46, opened 2021-12-17
Add impl to generate child public key from parent public key when not hardened


satran004:

zack-scott:
I have implemented some of the logic to start working on this, and I am getting the derivation for the chain code to come out correct. But, I am not having luck with getting the public key to come out correct. I am sure I am doing something wrong with the math somewhere.

`    /**
     * Derive the public child key from HD parent public key
     *
     * @param parent     the parent key
     * @param child      the child index
     * @return
     */
    public HdPublicKey getPublicChildKey(HdPublicKey parent, long child) {
        HdPublicKey publicKey = new HdPublicKey();

        byte[] childNumber = BytesUtil.ser32LE(child);


        publicKey.setVersion(parent.getVersion());
        publicKey.setDepth(parent.getDepth() + 1);
        publicKey.setChildNumber(childNumber);


        //If derivation V2 Shelley
        byte[] AP = parent.getKeyData();
        byte[] cP = parent.getChainCode();

        byte[] I;
    
        byte[] data = BytesUtil.merge(new byte[]{2}, AP, childNumber);
        byte[] IL = Arrays.copyOfRange(Hmac.hmac512(data, cP), 0, 28);
        
        data[0] = 3;
        I = Hmac.hmac512(data, cP);
        
        byte[] IR = Arrays.copyOfRange(I, 32, 64);
        publicKey.setChainCode(IR); // verified to be correct chain code




        BigInteger kLiBI = parseUnsignedLE(IL)
                .multiply(BigInteger.valueOf(8))
                .add(parseUnsignedLE(AP));
        BigInteger order = BigInteger.valueOf(2).pow(252)
                .add(new BigInteger("27742317777372353535851937790883648493"));
        if (kLiBI.mod(order).equals(BigInteger.ZERO)) {
            return null;
        }
        IL = serializeUnsignedLE256(kLiBI);

        byte[] A = ED25519SPEC.getB().scalarMultiply(IL).toByteArray();

        publicKey.setKeyData(A);



        return publicKey;
    }`

satran004:
@zack-scott Thanks for trying this out. 

I had faced some issue while trying to implement this earlier. 
As I am traveling now, I have limited access to my laptop for next few days.
I will try to verify your impl with the spec next week.

zack-scott:
[...]

PR #128, opened 2022-06-23
Implemented getChildPublicKey for xPub key derivation

zack-scott:
Added the function and test mentioned in https://github.com/bloxbean/cardano-client-lib/issues/46 to the code for a pull request

### DIFF
--- a/src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
+++ b/src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
@@ -215,7 +215,7 @@ public class HdKeyGenerator {
      * @param child      the child index
      * @return
      */
-    public HdPublicKey getChildPublicKey(HdPublicKey parent, long child) {
+    public HdPublicKey getChildPublicKey(HdPublicKey parent, int child) {
         HdPublicKey publicKey = new HdPublicKey();
         byte[] AP = parent.getKeyData();
 
--- a/src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
+++ b/src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
@@ -7,11 +7,16 @@ import com.bloxbean.cardano.client.crypto.bip32.util.Hmac;
 import com.bloxbean.cardano.client.crypto.cip1852.DerivationPath;
 import com.bloxbean.cardano.client.util.HexUtil;
 import com.bloxbean.cardano.client.util.OSUtil;
+
+import net.i2p.crypto.eddsa.math.Curve;
+import net.i2p.crypto.eddsa.math.GroupElement;
+import net.i2p.crypto.eddsa.math.FieldElement;
 import net.i2p.crypto.eddsa.spec.EdDSANamedCurveTable;
 import net.i2p.crypto.eddsa.spec.EdDSAParameterSpec;
 import org.bouncycastle.crypto.digests.SHA512Digest;
 import org.bouncycastle.crypto.generators.PKCS5S2ParametersGenerator;
 import org.bouncycastle.crypto.params.KeyParameter;
+import org.bouncycastle.math.ec.custom.djb.Curve25519;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -21,6 +26,7 @@ import javax.crypto.spec.PBEKeySpec;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
 import java.security.NoSuchAlgorithmException;
+import java.security.PublicKey;
 import java.security.spec.InvalidKeySpecException;
 import java.util.Arrays;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
```diff
@@ -7,11 +7,16 @@
...
+
+import net.i2p.crypto.eddsa.math.Curve;
+import net.i2p.crypto.eddsa.math.GroupElement;
+import net.i2p.crypto.eddsa.math.FieldElement;
 import net.i2p.crypto.eddsa.spec.EdDSANamedCurveTable;
 import net.i2p.crypto.eddsa.spec.EdDSAParameterSpec;
 import org.bouncycastle.crypto.digests.SHA512Digest;
 import org.bouncycastle.crypto.generators.PKCS5S2ParametersGenerator;
 import org.bouncycastle.crypto.params.KeyParameter;
+import org.bouncycastle.math.ec.custom.djb.Curve25519;
```
satran004, line 19: Please remove unused imports

src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
```diff
@@ -206,6 +212,58 @@ public HdKeyPair getChildKeyPair(HdKeyPair parent, long child, boolean isHardene
...
     }
 
+    /**
+     * Derive the public child key from HD parent public key
+     *
+     * @param parent     the parent key
+     * @param child      the child index
+     * @return
+     */
+    public HdPublicKey getChildPublicKey(HdPublicKey parent, long child) {
```
satran004, line 222: Can we use "int" type instead of "long" for child parameter ?
As per the spec, for soft index =  i < 2^31. For hardened index, i ≥ 2^31

As pub key -> pub key derivation is only applicable for soft index, I think we should be ok with "int" type as the max value of int is 2,147,483,648 (2 ^31)

src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
```diff
@@ -206,6 +212,58 @@ public HdKeyPair getChildKeyPair(HdKeyPair parent, long child, boolean isHardene
...
+        //Ai ← AP + [8ZL]B,
+        BigInteger kLiBI = parseUnsignedLE(ZL)
+                .multiply(BigInteger.valueOf(8));
+
+        byte[] kLi = serializeUnsignedLE256(kLiBI);
+
+        GroupElement gp1 = new GroupElement(ED25519SPEC.getCurve(), AP);
+        gp1 = gp1.toCached();
+        GroupElement groupElement = ED25519SPEC.getB().scalarMultiply(kLi).add(gp1);
+        //TODO -- If Ai is the identity point (0, 1), discard the child
```
satran004, line 254: If possible, please do this check. Otherwise, you can leave that for now.

src/test/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGeneratorTest.java
```diff
@@ -49,4 +50,18 @@ void testGetRootKeyPairFromEntropy_whenAndroid() throws MnemonicException.Mnemon
         assertThat(rootKey).isEqualTo("root_xsk1hp9an83kfma0ufdaeqft6xv0snf4ek9uqemk5chp3p25el8w0fglszmgkq9qxvguj33fnulms4qfnx9jawhde4ng9qzg3zzg5u8r2af4jxpe8nfjulrzey7p8ttnt5yn53exsawm6wkmtqm989ehwtr0kc244zfn");
 
     }
+
+    @Test
+    void testPubKeyFromParentPubKey() {
```
satran004, line 55: Should we add some more tests? Please check if any test vector available for bip32-ed25519. Otherwise, you can ignore that for now. I will check after merge.

zack-scott:
I have removed unused libraries in the code. I also switched long to int for the child field in the newly added function @satran004 .

src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
zack-scott: I will need to do some research into the identity point.

src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
zack-scott: I have switched to int child.

src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
zack-scott: I have removed unused imports

src/main/java/com/bloxbean/cardano/client/crypto/bip32/HdKeyGenerator.java
satran004: Thanks. I will merge this PR for now. You can impl identity point check through another PR.

PR #150, opened 2022-08-11
Replace Ogmios Backend implementation with Ogmios Java Client Library Dependency from adabox.io

edridudi:


satran004 (approved):
Thanks @edridudi for the integration.

PR #158, opened 2022-09-30
[WIP] Added new networks

nejcr:
I added new networks, however, I am not deeply familiar with this lib, and if someone (@satran004) could provide me the correct integer values I will update it accordingly.

### DIFF
--- a/src/main/java/com/bloxbean/cardano/client/address/util/AddressEncoderDecoderUtil.java
+++ b/src/main/java/com/bloxbean/cardano/client/address/util/AddressEncoderDecoderUtil.java
@@ -34,6 +34,8 @@ public class AddressEncoderDecoderUtil {
         String prefixTail;
         switch (network) {
             case TESTNET:
+            case PREVIEW:
+            case PREPROD:
                 prefixTail = "_test";
                 break;
             case MAINNET:
@@ -52,6 +54,10 @@ public class AddressEncoderDecoderUtil {
             network = NetworkId.MAINNET;
         } else if (Networks.testnet().equals(networkInfo)) {
             network = NetworkId.TESTNET;
+        } else if (Networks.preview().equals(networkInfo)) {
+            network = NetworkId.PREVIEW;
+        } else if (Networks.preprod().equals(networkInfo)) {
+            network = NetworkId.PREPROD;
         } else {
             throw new AddressRuntimeException("Unknown network type - " + networkInfo);
         }
--- a/src/main/java/com/bloxbean/cardano/client/common/model/Networks.java
+++ b/src/main/java/com/bloxbean/cardano/client/common/model/Networks.java
@@ -10,4 +10,14 @@ public class Networks {
         Network testnet = new Network(0b0000, 1097911063);
         return testnet;
     }
+
+    public static Network preprod() {
+        Network testnet = new Network(0b0000, 1);
+        return testnet;
+    }
+
+    public static Network preview() {
+        Network testnet = new Network(0b0000, 2);
+        return testnet;
+    }
 }
--- a/src/main/java/com/bloxbean/cardano/client/transaction/spec/NetworkId.java
+++ b/src/main/java/com/bloxbean/cardano/client/transaction/spec/NetworkId.java
@@ -2,5 +2,7 @@ package com.bloxbean.cardano.client.transaction.spec;
 
 public enum NetworkId {
     TESTNET,
-    MAINNET
+    MAINNET,
+    PREVIEW,
+    PREPROD
 }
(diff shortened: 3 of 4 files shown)


satran004:
@nejcr Thanks for working on this PR.

For new networks like prepod and preview,  networkId (also known as Network Tag) is still 0 as these are test networks.
AFAIK there are only two possible values for network id   1 (for mainnet) and 0 (for all testnets including custom private testnets).

You can refer to this CIP for more details https://cips.cardano.org/cips/cip19/#networktag

But, the Network class in cardano-client has an another field ``protocolMagic`` which is network specific.  Prepod = 1 and Preview = 2. But protocol magic is not used directly inside cardano client lib.

Network Id is used in Address generation and during transaction serialization (optional).

Please refer to my other comments inline.

src/main/java/com/bloxbean/cardano/client/address/util/AddressEncoderDecoderUtil.java
```diff
@@ -34,6 +34,8 @@ public static String getPrefixTail(NetworkId network) {
         String prefixTail;
         switch (network) {
             case TESTNET:
+            case PREVIEW:
```
satran004, line 37: We don't need PREVIEW and PREPOD here. Only TESTNET networkId is enough.  So NetworkId enum will have two values TESTNET, MAINNET where TESTNET is for all testnets

src/main/java/com/bloxbean/cardano/client/address/util/AddressEncoderDecoderUtil.java
```diff
@@ -52,6 +54,10 @@ public static NetworkId getNetworkId(Network networkInfo) {
             network = NetworkId.MAINNET;
```
satran004, line 54: I think we need to change this logic. In existing code, Network object is being compared but instead we should compare NetworkId of the network and return NetworkId.MAINNET or NetworkId.TESTNET


Some thing like
```
if (NetworkId.MAINNET.equals(networkInfo.getNetworkId()) return NetworkId.MAINNET;
else  if //check for network TESTNET network id

else throw AddressRuntimeException
```

src/main/java/com/bloxbean/cardano/client/common/model/Networks.java
```diff
@@ -10,4 +10,14 @@ public static Network testnet() {
         Network testnet = new Network(0b0000, 1097911063);
         return testnet;
     }
+
```
satran004, line 13: This looks good to me. You may want to change the local variable name in prepod & preview methods to prepod/preview instead of testnet.

src/main/java/com/bloxbean/cardano/client/transaction/spec/NetworkId.java
```diff
@@ -2,5 +2,7 @@
 
 public enum NetworkId {
     TESTNET,
-    MAINNET
+    MAINNET,
+    PREVIEW,
```
satran004, line 6: PREVIEW and PREPOD are not required as only two possible values for NetworkId.

src/main/java/com/bloxbean/cardano/client/transaction/spec/TransactionBody.java
```diff
@@ -155,6 +155,14 @@ public Map serialize() throws CborSerializationException, AddressExcepion {
                 case TESTNET:
                     bodyMap.put(new UnsignedInteger(15), new UnsignedInteger(0));
                     break;
+                case PREVIEW:
```
satran004, line 158: These are network Ids. So only 1, 0.  This change is not required.

src/main/java/com/bloxbean/cardano/client/transaction/spec/TransactionBody.java
```diff
@@ -298,7 +306,15 @@ public static TransactionBody deserialize(Map bodyMap) throws CborDeserializatio
             int networkIdInt = networkIdUI.getValue().intValue();
             if (networkIdInt == 0) {
                 transactionBody.setNetworkId(NetworkId.TESTNET);
-            }else if (networkIdInt == 1) {
+            }
```
satran004, line 309: Same as above. This change is not required.

satran004 (changes requested):
@nejcr Can you please also add few tests to include PREVIEW, PREPOD network ?

Basically we need to check if a developer pass the network as PREVIEW or PREPOD, the generated address should be similar to regular testnet (starts with addr_test). I think we can add few tests in AddressTest class or AccountTest class.

satran004:
@nejcr Let me know if you need any additional info for this PR

satran004:
Thanks @nejcr for the changes. Looks good to me. Merging it to a temp branch for now. 

I will do a minor change to compare networkid instead of network obj in ``AddressEncoderDecoderUtil`` and merge back to master.

PR #168, opened 2022-10-14
Static unit Plutus data (often used for redeemers)

will991:
Proposing a unit constructor for redeemers.


satran004:
Thanks @will991 . Makes sense to me. 
Just wondering if we add this to ConstrPlutusData class. Any reason why it's in PlutusData ?

will991:
I thought since `unit` returns `PlutusData` and the expected parameter type for `Redeemer#data` expects `PlutusData`, it makes `PlutusData` the most obvious place where I would expect it when exploring the API.

satran004:
Thanks @will991. Merging it

PR #175, opened 2022-10-29
Royalty Tokens (CIP-27) Support

edridudi:


src/main/java/com/bloxbean/cardano/client/metadata/cbor/MetadataHelper.java
```diff
@@ -50,9 +50,13 @@ public static DataItem objectToDataItem(Object value) {
         }
     }
 
-    public static void checkLength(String str) {
```
satran004, line 53: Can you please add a test to cover this new change

src/main/java/com/bloxbean/cardano/client/util/JsonUtil.java
```diff
@@ -32,4 +33,8 @@ public static String getPrettyJson(String jsonStr) {
     public static JsonNode parseJson(String jsonContent) throws JsonProcessingException {
         return mapper.readTree(jsonContent);
     }
+
+    public static String[] usingSplitMethod(String text, int n) {
```
satran004, line 37: If I understand correctly, this method is not specific to json. Should we create a new util for string operations and add to that? Also please add a test for this new method

Issue #181, opened 2022-11-18
wrong scriptDataHash for no-redeemers but datum in withnessSet


sacrelege:
In case a datum is attached to the witnessSet but no redeemers the scriptDataHash needs to be calculated as follows:

Bytes.concat( HexUtil.decodeHexString("0x80"), plutusDataBytes, HexUtil.decodeHexString("0xA0"));

ref: https://github.com/input-output-hk/cardano-ledger/blob/8b6f8e1a75034ca66fd66a39d437252eec927d71/eras/babbage/test-suite/cddl-files/babbage.cddl#L134

example

datum hash: b5c5a700e5b7f1664c689caf8293fbbe199e2ebbe1857394e0a2098eed00dd81
datum hex:
d8799f4114d8799fd8799fd8799fd8799f581c3050f6f4d5981748bc3a2b84d8165b20c100a75057b6593befd9323cffd8799fd8799fd8799f581cc5cdc99429b4ce659f2542994c48b6c801f0b8e21ca7fb586326a545ffffffffd87a80ffd87a80ff1a002625a0d8799fd879801a0025a559d8799f01ffffff
scriptDataHash=[71, -22, -92, 74, -39, 124, 55, -108, 120, -127, -125, 119, 41, -77, 48, -72, 121, 0, -10, -77, -29, 103, -99, -9, -111, -118, 11, -126, -52, -29, -81, 105]

PR #182, opened 2022-11-20
[#181] Fix - Changed the generation logic for empty redeemers and non…

satran004:
- Updated script datahash generation logic for empty redeemers and non empty datum

```
; Finally, note that in the case that a transaction includes datums but does not
; include any redeemers, the script data format becomes (in hex):
; [ 80 | datums | A0 ]
; corresponding to a CBOR empty list and an empty map.
```

https://github.com/input-output-hk/cardano-ledger/blob/8b6f8e1a75034ca66fd66a39d437252eec927d71/eras/babbage/test-suite/cddl-files/babbage.cddl#L134


satran004:
Fyi. @sacrelege

PR #192, opened 2022-12-17
Enhancements and Koios Bump to 1.16.1

edridudi:
- [New Tokens Distribution Guide + Example](https://github.com/bloxbean/cardano-client-lib/commit/b0a4320e7957cd4dac651e3d9ac444b9a3c5fa5a)
- [NFT Metadata Enhancement to Support a metadata field value for Map values of type List](https://github.com/bloxbean/cardano-client-lib/commit/1d5a2697d946d4b40f0bef32c5b263db9b7c3e24)
- [Add New Services Methods: getAllAccountAddresses, getAllAccountAssets, getAllAssetAddresses, getAllPolicyAssets](https://github.com/bloxbean/cardano-client-lib/commit/0ab0324239c298f2ec338e05fa610bf625650537)
- [Bump Koios Java Client to 1.16.1 (TransactionContent Mapping Fix)](https://github.com/bloxbean/cardano-client-lib/commit/fbc53b8a8b5ecd49dc493997e8f54a4484eb3080)

### DIFF
--- a/backend-modules/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosAccountService.java
+++ b/backend-modules/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosAccountService.java
@@ -149,6 +149,27 @@ public class KoiosAccountService implements com.bloxbean.cardano.client.backend.
     }
 
     @Override
+    public Result<List<AccountAddress>> getAllAccountAddresses(String stakeAddress) throws ApiException {
+        List<AccountAddress> accountAddresses = new ArrayList<>();
+        int page = 1;
+        Result<List<AccountAddress>> accountAddressesResult = getAccountAddresses(stakeAddress, 1000, page);
+        while (accountAddressesResult.isSuccessful()) {
+            accountAddresses.addAll(accountAddressesResult.getValue());
+            if (accountAddressesResult.getValue().size() != 1000) {
+                break;
+            } else {
+                page++;
+                accountAddressesResult = getAccountAddresses(stakeAddress, 1000, page);
+            }
+        }
+        if (!accountAddressesResult.isSuccessful()) {
+            return accountAddressesResult;
+        } else {
+            return Result.success(accountAddressesResult.toString()).withValue(accountAddresses).code(accountAddressesResult.code());
+        }
+    }
+
+    @Override
     public Result<List<AccountAddress>> getAccountAddresses(String stakeAddress, int count, int page) throws ApiException {
         return this.getAccountAddresses(stakeAddress, count, page, OrderEnum.asc);
     }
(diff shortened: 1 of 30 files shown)


edridudi:
Fixes #193

backend-modules/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosAccountService.java
```diff
@@ -148,6 +148,27 @@ private Result<List<AccountHistory>> convertToAccountHistories(List<AccountHisto
         return Result.success("OK").withValue(accountHistoryList).code(200);
     }
 
+    @Override
+    public Result<List<AccountAddress>> getAllAccountAddresses(String stakeAddress) throws ApiException {
+        List<AccountAddress> accountAddresses = new ArrayList<>();
+        int page = 1;
+        Result<List<AccountAddress>> accountAddressesResult = getAccountAddresses(stakeAddress, 1000, page);
```
satran004, line 155: For BF impl, the page size is 100. But for Koios it's 1000. Any reason why it's not 100?

cip/cip25/src/main/java/com/bloxbean/cardano/client/cip/cip25/NFTProperties.java
```diff
@@ -44,22 +44,30 @@ public String getProperty(String name) {
         return (String) get(name);
     }
 
-    public NFTProperties property(String name, java.util.Map<String, String> values) {
+    public NFTProperties property(String name, java.util.Map<String, Object> values) {
```
satran004, line 47: Can you please add javadoc with info like kind of values expected? String and List etc.

satran004: Also, if the value is not String or List, should we throw IllegalArgumentException instead of ignoring it?

satran004 (commented):
Thanks @edridudi for the PR. Few comments. Otherwise looks good to me

backend-modules/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosAccountService.java
edridudi: Koios Max Page Size is 1000, for BF is 100

cip/cip25/src/main/java/com/bloxbean/cardano/client/cip/cip25/NFTProperties.java
edridudi: > Can you please add javadoc with info like kind of values expected? String and List etc.

Sure.
Values expected are only String or List<String>

> Also, if the value is not String or List, should we throw IllegalArgumentException instead of ignoring it?

Sure good idea!

satran004:
Thanks. Merging. 
I will move Utxo.deserialize() method to a separate util class in cip30 module after merge as the it's cip30 specific cbor.

PR #208, opened 2023-01-22
Koios Java Client Bump to 1.16.2 + CIP30 Value Serialize/Deserialize

edridudi:


satran004:
@edridudi Please check the build error

cip/cip30/src/main/java/com/bloxbean/cardano/client/cip/cip30/CIP30ValueDeserializer.java
```diff
@@ -0,0 +1,48 @@
...
+    private CIP30ValueDeserializer() {
+        throw new IllegalStateException("Utility Class");
+    }
+
+    /**
+     * Deserialize cbor from CIP30's wallet.getBalance() method
+     * @param bytes balance cbor bytes from wallet.getBalance()
+     * @return {@link Value}
+     */
+    public static Value deserialize(byte[] bytes) throws CborDeserializationException {
```
satran004, line 25: @edridudi If we move Value's deserialization code from TransactionOutput to Value class, do we still need this class? I understand why you have introduced this class now, but it may make sense to keep this logic in one place for easy maintenance in future. Because now there are 3 different ways value can be represented between Shelley to Babbage. For any change in future, it has to be done in both places.

cip/cip30/src/main/java/com/bloxbean/cardano/client/cip/cip30/CIP30ValueSerializer.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.math.BigInteger;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Serialize Value Object to cbor CIP30's Balance Object
+ * to Support CIP30's wallet.getUtxo(amount: cbor&lt;value&gt;) method
+ */
+public class CIP30ValueSerializer {
```
satran004, line 18: Same as CIP30Deserializer, do we need this class ? We can keep this class but it will invoke Value class.

edridudi: We don't need these Classes as long as the serialization & deserialization functions have the same functionality as in these classes.
I agree that we should have them in one single place. that would be even better.

edridudi:
@satran004 ,
Updated PR with CIP30ValueDeserializer.java, CIP30ValueSerializer.java, CIP30ValueTest.java Removed.
Moved CIP30ValueTest Unit-Tests to ValueSpecTest (Errors in this PR will be resolved after merging with #210)

satran004:
@edridudi PR #210 is merged now. Can you please update your repo and remove CIP30ValueDeserializer.java ? Thanks

PR #215, opened 2023-02-02
Many Fixes :)

edridudi:


edridudi:
Not sure about Sonatype's suggested changes.
Looks like we should ignore them.

function/src/main/java/com/bloxbean/cardano/client/function/TxOutputBuilder.java
```diff
@@ -23,6 +23,7 @@ default TxOutputBuilder and(TxOutputBuilder after) {
         };
     }
 
+    @Deprecated
     default TxBuilder buildInputs(TxInputBuilder after) {
```
satran004, line 27: @edridudi Let's not deprecate the current buildInputs(TxBuilderInput), it should just call the new buildInput() with separateChangeOutput = false. So the default behavior is still same. Also, please add some javadoc.

function/src/main/java/com/bloxbean/cardano/client/function/TxOutputBuilder.java
```diff
@@ -23,40 +23,26 @@ default TxOutputBuilder and(TxOutputBuilder after) {
...
-                        txOutputSameAddress.setValue(txOutputSameAddress.getValue().plus(txOutput.getValue()));
-                    } else {
-                        transaction.getBody().getOutputs().add(txOutput);
-                    }
-                });
-
-            //Clear multiasset in the context
-            context.clearMintMultiAssets();
-        };
+        return buildInputs(after, false);
```
edridudi, line 34: it's a mistake here we should send it with true not false.

satran004: @edridudi I have updated separateChangeOutput to mergeChangeOutput

https://github.com/bloxbean/cardano-client-lib/commit/99e4dd5dbcf6fb57a789b50349dafd5d8d86ce7b

edridudi: Ok cool

PR #236, opened 2023-04-09
Alignment with Koios Java Client 1.17.1 + New Get Transactions API

edridudi:


integration-test/src/it/java/com/bloxbean/cardano/client/backend/api/TransactionServiceIT.java
```diff
@@ -44,6 +44,15 @@ public void testGetTransaction() throws Exception {
         System.out.println(JsonUtil.getPrettyJson(result.getValue()));
     }
 
+    @Test
+    public void testGetTransactions() throws Exception {
+        String txnHash = "fb77161efb9788f78d13e815e7c0bfde192cbdfb18a8e502c7a3d3171c1e5688";
+        Result<List<TransactionContent>> result = service.getTransactions(List.of(txnHash));
+
+        assertNotNull(result.getValue());
```
satran004, line 52: @edridudi Thanks. The changes look good to me. 
Can you please update the assert statement to check if the return value has the same tx content/id? Please also add one negative test .

satran004 (changes requested):

integration-test/src/it/java/com/bloxbean/cardano/client/backend/api/TransactionServiceIT.java
edridudi: @satran004 What kind of negative test?

satran004: @edridudi Like if one of the txId in txIdList is not valid, it should return error instead of partial or empty list. That's according to the current impl.

Issue #224, opened 2023-03-15
Split core module to serialization and plutus annotation


satran004:
Currently, core module contains classes with different functionalities. There is an opportunity to split core into multiple modules
1.  serialization/spec/or a better name (Module with serialization logic)
2.  Plutus annotation (Module with plutus annotations and converters). We may call this module plutus-helper/plutus-extra to increase the scope of this module.
3. core (Remaining functionalities)

Note:
- As serialization code is used everywhere in CCL and also application code, we should not change the package name in ```serialization``` module.

PR #242, opened 2023-04-24
reafactor(core) : #224 Split core module into spec, plutus, coinselec…

satran004:
This PR is for #224 

The Core module is divided into multiple sub-modules. Each of these sub-modules can now be added independently.

   1. **Spec:** This module contains the low-level CBOR serialization logic and a few utilities. To build a transaction using the low-level API, this module should be enough.

   2. **Plutus:** This module contains Plutus annotations and helpers that can be used to convert/serialize a POJO to PlutusData and vice versa.

   3. **CoinSelection:** This module contains the CoinSelection API and its implementations.

   4. **Core-API:** This module contains the core APIs required for the transaction builder or other high-level modules, such as UtxoSupplier, ProtocolParamSupplier, Utxo, Amount, etc

   5. **Core:**  This module contains the remaining classes for legacy reasons, such as legacy high-level APIs. The legacy high-level APIs will be removed once we have the new high-level APIs later.
   
   **Known issues:** 
   - Some of the unit tests are still in old core module due to some dependencies. This will be fixed later. 
   - Also, few packages are not consistent with new module names. This will be fixed later with deprecation notice.


matiwinnetou:
reafactor -> refactor.

Once this is there, I will use new library as part of merkle-tree-java project - removing core and putting plutus one from here.

Issue #301, opened 2023-08-23
QuickTX collateral selection results in InsufficientCollateral


stikkos:
I'm trying to send a scriptTx using QuickTx but getting: "InsufficientCollateral (Coin (-12084272)) (Coin 307193)"

Collateral summary:
IN
    5f7e7cae5ac68122fd07ae76977034802977c160e5d015cbeeb80db2e4b618d2 - index 0 - 9793445 lovelace
    3e2bdf7f96ff13d552bd7a4979b04f710665fca59b98af8e1461a0fe31475c97 - index 1 - 2598020 lovelace    

OUT collateralReturn:   12084272
Total collateral:              307193

Error:
    InsufficientCollateral (Coin (-12084272)) (Coin 307193)

The first issue is that, even though an exact UTXO for the `DEFAULT_COLLATERAL_AMT` is present, the UTXO selection strategy doesn't always select this one.
Additionally, the calculated total collateral seems to be insufficient...

- maybe good to use a dedicated "fixed value" UTXO selection strategy first with the aim of selecting an UTXO containing exact 5 ADA before using default UTXO selection strategy
- it would be helpful to add functionality to QuickTx allowing users to specify collateral UTXO (hash + index) explicitly. When set, this would have priority over UTXO selection (and can be used as-is, i.e without changes based on CollateralPercentage)

Full Transaction (pre-prod):

```
{
  "body" : {
    "inputs" : [ {
      "transactionId" : "f02f73f6732020a2251d47360809a1475e4401f359ce8d7faed42479711b49a4",
      "index" : 0
    } ],
    "outputs" : [ {
      "address" : "addr_test1qrqhq0xp6ach453pm5cfjmhg3jg68tcvlgn80nv8zsfth956mxp5gzhjqj8xj9zdlrg6v6nr9jqwpzrvf0vmmrhuztdstmsl7l",
      "value" : {
        "coin" : 9795205,
        "multiAssets" : [ ]
      },
      "datumHash" : null,
      "inlineDatum" : null,
      "scriptRef" : null
    } ],
    "fee" : 204795,
    "ttl" : 37064850,
    "certs" : [ ],
    "withdrawals" : [ ],
    "update" : null,
    "auxiliaryDataHash" : "FrS36GUFCrqmie6wVQvEh9Grvzo82xluIuOeHzhOzmE=",
    "validityStartInterval" : 0,
    "mint" : [ ],
[...]

PR #306, opened 2023-08-26
feat: #301 Added withCollateralInputs() to add specific collateral in…

satran004:
New method to add specific collaterals

- withCollateralInputs(TransactionInput... inputs) 

- Wrapper UtxoSelectionStrategy and UtxoSelector to exclude specific collateral from coin selection


satran004:
@stikkos fyi.

PR #312, opened 2023-08-30
Added endpoints

Thomar10:


satran004 (changes requested):
@Thomar10 Thanks a lot for the PR. It looks good to me. Just few comments.

Can you please  also add integration tests for these two new methods? Blockfrost related integration tests can be found in "integration-test" module.

PR #315, opened 2023-09-01
Added AssetService.getTransactions, added Tuple.toString + equals & hashCode, change AddressProvider visibility

stikkos:
- added `getTransactions` to `AssetService` (https://docs.blockfrost.io/#tag/Cardano-Assets/paths/~1assets~1%7Basset%7D~1transactions/get)
	this method is particularly useful for efficiently fetching DEX pool txs

- add `@ToString` and `@EqualsAndHashCode` to `com.bloxbean.cardano.client.util.Tuple`
	for easier debugging and comparing

- changed visibility of 1 `AddressProvider` method
	for easier conversion of stake and payment key hash to address

### DIFF
--- a/address/src/main/java/com/bloxbean/cardano/client/address/AddressProvider.java
+++ b/address/src/main/java/com/bloxbean/cardano/client/address/AddressProvider.java
@@ -375,7 +375,7 @@ public class AddressProvider {
         }
     }
 
-    public static Address getAddress(byte[] paymentKeyHash, byte[] stakeKeyHash, byte headerKind, Network networkInfo, AddressType addressType) {
+    private static Address getAddress(byte[] paymentKeyHash, byte[] stakeKeyHash, byte headerKind, Network networkInfo, AddressType addressType) {
         NetworkId network = getNetworkId(networkInfo);
 
         //get prefix
--- a/address/src/main/java/com/bloxbean/cardano/client/address/AddressProvider.java
+++ b/address/src/main/java/com/bloxbean/cardano/client/address/AddressProvider.java
@@ -388,7 +388,7 @@ public class AddressProvider {
         return new Address(prefix, addressArray);
     }
 
-    public static byte[] getAddressBytes(byte[] paymentKeyHash, byte[] stakeKeyHash, AddressType addressType, byte header) {
+    private static byte[] getAddressBytes(byte[] paymentKeyHash, byte[] stakeKeyHash, AddressType addressType, byte header) {
         //get body
         byte[] addressArray;
         switch (addressType) {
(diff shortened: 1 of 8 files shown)


address/src/main/java/com/bloxbean/cardano/client/address/AddressProvider.java
```diff
@@ -375,7 +375,7 @@ public static Address getRewardAddress(@NonNull Credential stakeCredential, Netw
         }
     }
 
-    private static Address getAddress(byte[] paymentKeyHash, byte[] stakeKeyHash, byte headerKind, Network networkInfo, AddressType addressType) {
+    public static Address getAddress(byte[] paymentKeyHash, byte[] stakeKeyHash, byte headerKind, Network networkInfo, AddressType addressType) {
```
satran004, line 378: @stikkos Rather than exposing this method, could you use the following method, which accepts Credential objects as parameters? I think that this method is too low-level, as developers might pass an incorrect headerKind byte.

```
Address getBaseAddress(Credential paymentCredential, Credential delegationCredential, Network networkInfo) 
```

Similar methods exist for creating Enterprise  and Reward addresses.

backend-modules/blockfrost/src/main/java/com/bloxbean/cardano/client/backend/blockfrost/service/BFAssetService.java
```diff
@@ -119,4 +120,23 @@ public Result<List<PolicyAsset>> getPolicyAssets(String policyId, int count, int
...
+    @Override
+    public Result<List<AssetTransactionContent>> getTransactions(String unit, int count, int page, OrderEnum order) throws ApiException {
+        if(order == null)
+            order = OrderEnum.asc;
+
+        Call<List<AssetTransactionContent>> call = assetsApi.getTransactions(getProjectId(), unit, count, page, order.toString());
+
+        try {
+            Response<List<AssetTransactionContent>> response = call.execute();
+            if (response.isSuccessful())
```
satran004, line 133: I think we can replace these two lines with processResponse method call

```
processResponse(response) 
```

satran004 (changes requested):
Thanks a lot @stikkos  for the PR.

Few small feedback. Please check.

address/src/main/java/com/bloxbean/cardano/client/address/AddressProvider.java
stikkos: Yes, great. I can use this method indeed. Have reverted the visibility change

backend-modules/blockfrost/src/main/java/com/bloxbean/cardano/client/backend/blockfrost/service/BFAssetService.java
stikkos: Have adjusted

Issue #337, opened 2023-10-18
Add support for plain vanilla transaction signing.


matiwinnetou:
Transaction deserialisation between various Cardano clients can be tricky so if one client prepares a transaction and CCL deserialises it and we sign it, it is not guaranteed to work. 

In fact we recently stumbled exactly this case with Hydra, where new Hydra 0.13.x protocol prepares transactions which are not sign to be signed by the client. Signing them with CCL using Transaction.deserialisation and TxSigner fails.

Provided ability in quick-tx and outside of quick-tx to simply sign transactions based on bytes only, without having to deserialise, sign and then serialise again the whole transaction.

PR #342, opened 2023-11-06
fix: #337 Added a separate method to sign transaction bytes

satran004:
- New methods have been added that avoid de-serializing the full transaction bytes, thereby preventing any compatibility issues.

```
 public byte[] sign(byte[] transactionBytes, SecretKey secretKey)
public byte[] sign(byte[] transactionBytes, HDKeyPair hdKeyPair)
```

- Old sign methods are delegating to above new methods for tx signing


satran004:
@Sotatek-HuyLe3 Please check TransactionBytes class in this PR. We can use this class or something like this in Yaci to fix incorrect metadata cbor.

Sotatek-HuyLe3a:
Thanks, this class could be  useful for fixing incorrect metadata cbor.

Kammerlo (approved):

PR #346, opened 2023-11-10
Koios Java Client Aligment to v1.18.1 + Fix #50 + Fix #64

### LINKED ISSUE #50: Please add "addresses/{address}/total" support
As far as I investigated, I have not found the implementation for the "addresses/{address}/total" support, bloxbean version 0.1.5.

I believe, that this addition will help users like me to get full information regarding address without a need to perform additional requests to the API.

Please kindly refer to the provided screenshot for more information.
image

### LINKED ISSUE #64: Fetch SLOTS_PER_EPOCH Field from NetworkParameters instead of hardcoded value in PolicyUtil Class.

edridudi:


backend-modules/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosEpochService.java
```diff
@@ -5,27 +5,75 @@
...
+import java.util.*;
+import java.util.concurrent.atomic.AtomicInteger;
 
 /**
  * Koios Epoch Service
  */
 public class KoiosEpochService implements EpochService {
 
+    private final ObjectMapper objectMapper = new ObjectMapper();
+    private final static String PLUTUS_COST_MODELS_URL = "https://raw.githubusercontent.com/input-output-hk/plutus/master/plutus-core/cost-model/data/builtinCostModel.json";
```
satran004, line 29: @edridudi Do we really need this dependency? As discussed earlier, it can just rely on operations defined in the PlutusOps class, and we will update this class whenever there is a new Plutus version. 

I don't think we should depend on a file from an external repository, as this code will break if the file is moved to a different location.

satran004 (changes requested):
Thanks @edridudi. Just one comment.

backend-modules/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosEpochService.java
edridudi: it's there by mistake I will remove

PR #350, opened 2024-01-07
Implement cip 67

Kammerlo:


cip/cip67/src/main/java/com/bloxbean/cardano/client/cip/cip67/CIP67.java
```diff
@@ -0,0 +1,82 @@
+package com.bloxbean.cardano.client.cip.cip67;
+
+import com.bloxbean.cardano.client.crypto.CRC8;
+import com.bloxbean.cardano.client.util.IntUtil;
+
+import java.nio.ByteBuffer;
+
+public class CIP67 {
```
satran004, line 8: Thanks @Kammerlo 

Should we rename CIP67 to CIP67AssetName or CIP67AssetNameUtil to give some context ?

common/src/main/java/com/bloxbean/cardano/client/util/IntUtil.java
```diff
@@ -0,0 +1,20 @@
+package com.bloxbean.cardano.client.util;
+
+public class IntUtil {
```
satran004, line 3: There is a Utils class which has a method with similar functionality ``uint32ToByteArrayBE``. But it's currently in crypto/bip32 package.

I am just wondering if we can use that. Otherwise, we can keep this class but rename it to ByteUtil, allowing us to add additional conversion methods, such as long to bytes, etc later.

Kammerlo: So yes the function would work as well. For me this conversions seemed pretty common so I wasn't looking into the bip32 package for this functionality. 
I renamed the package. We can discuss if moving the other functions to a more common place.

satran004 (approved):

PR #352, opened 2024-01-10
351 add support for metdata standard from cip68

Kammerlo:


plutus/src/main/java/com/bloxbean/cardano/client/plutus/spec/serializers/PlutusDataJsonConverter.java
```diff
@@ -63,4 +69,94 @@ else if (jsonNode.has(LIST)) {
...
-}
+
+    /**
+     * Convert {@link PlutusData} to utf8 encoded json String
+     * @param plutusData data to be encoded
+     * @return utf8 encoded json
+     * @throws JsonProcessingException
+     * @throws CborSerializationException
+     */
+    public static String toUTF8Json(PlutusData plutusData) throws JsonProcessingException, CborSerializationException {
```
satran004, line 80: This is specific to metadata used in PlutusData and may not be valid for regular Plutus data. I think it would be better to move this to another helper class or rename this method.

Can you move this method to another PR? I think we can then decide on the name, right package for this helper and add some more tests in that PR.

satran004 (commented):
Thanks @Kammerlo . Few minor comments.

cip/cip68/src/main/java/com/bloxbean/cardano/client/cip/cip68/CIP68TokenTemplate.java
```diff
@@ -26,41 +26,37 @@ public abstract class CIP68TokenTemplate<T extends CIP68TokenTemplate> extends N
...
-        this.assetName = assetName;
+        this.name = assetName;
     }
 
     public int getAssetNameLabel() {
         return assetNameLabel;
     }
 
-    public String getHexAssetFullName() {
+    public String getAssetNameAsBytes() {
```
satran004, line 36: @Kammerlo Let's just return byte[] . Also do we need to use name.getBytes(UTF-8) to convert to bytes to avoid any encoding issue?

plutus/src/test/java/com/bloxbean/cardano/client/plutus/spec/BytesPlutusDataTest.java
```diff
@@ -25,4 +28,15 @@ void serializeDeserialize() throws CborSerializationException, CborException, Cb
 
         assertThat(HexUtil.encodeHexString(bytes)).isEqualTo("4c48656c6c6f20576f726c6421");
     }
+
+    @Test
+    void unicodeDeserializeSerialize() throws CborDeserializationException, CborSerializationException {
+        UnicodeString unicodeString = new UnicodeString("Hello World!");
+        PlutusData deserialize = PlutusData.deserialize(unicodeString);
```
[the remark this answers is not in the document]
Kammerlo: Done. The test now validates if the new way of dezerializing a UnicodeString will be equal to directly create a BytesPlutusData of String.

satran004 (approved):

Issue #353, opened 2024-01-12
Wrong TxHash sporadically calculated by `TransactionUtil` class


nemo83:
There seems to be an issue that sporadically manifest in the method: 

`TransactionUtil.getTxHash(Transaction transaction)`

From time to time a wrong hash is created.

satran004:
@nemo83 
The Javadocs for these two methods highlight their differences.

There's a known issue where deserializing and serializing a transaction using different tools can result in varying byte sequences. CCL employs canonical ordering (e.g. in maps) during serialization. Therefore, if you serialize, deserialize, and then serialize a transaction created by CCL, you will consistently obtain the same result.

However, this may not hold true for transactions generated by other tools that do not adhere to canonical ordering. But, both types of transactions are valid during execution in node.

This is why it's advisable to avoid fully deserializing and serializing a transaction when calculating its Tx hash.

In the first method, ``TransactionUtil.getTransaction(Transaction)``, it is assumed that the transaction object is constructed by CCL. This is because it must undergo the serialize method of CCL, which orders keys in a map.

However, if you're using a transaction built by an external tool or an already executed transaction to obtain the Tx hash, it's better to use ``getTxHash(byte[] transactionBytes)``. This method doesn't involve full deserialization into a CCL object for Tx hash calculation. Instead, it deserializes to a DataItem and then serializes again with canonical order set to false, preserving the original order.

This method should be correct most of the time, but we've encountered issues with only few transactions on the mainnet when using this approach with Yaci. So in Yaci, we avoid deserializing to DataItem altogether. The parser simply extracts the transaction body bytes from the full transaction bytes to create the hash. 

I think we should update ``getTxHash(byte[] transaction)`` to adopt Yaci's approach. The other method, getTxHash(Transaction), should not be used for external or already executed transactions.

satran004:
@nemo83 
It's fixed now. PR https://github.com/bloxbean/cardano-client-lib/pull/356

Can you please verify the fix with your use case ? You can take the code from the below branch. Thanks

https://github.com/bloxbean/cardano-client-lib/tree/fix_353_main

nemo83:
Changes look good! Thank you

PR #356, opened 2024-01-14
fix: #353 Directly extract tx body bytes from tx bytes 

satran004:
#353


transaction-spec/src/test/java/com/bloxbean/cardano/client/transaction/util/TransactionUtilTest.java
```diff
@@ -67,4 +67,13 @@ public void getTxHash_fromBytes() {
...
+        var txBodyHex = HexUtil.encodeHexString(TransactionUtil.extractTransactionBodyFromTx(HexUtil.decodeHexString(cbor)));
+        System.out.println(txBodyHex);
```
Kammerlo, line 75: Just less than a minor thing: I think the System.out will just output a hex into the console which doesn't help the user at all. So it's not needed from my point of view, just to keep it clean.

Kammerlo (approved):

transaction-spec/src/test/java/com/bloxbean/cardano/client/transaction/util/TransactionUtilTest.java
satran004: Thanks. I will remove it.

nemo83:
Changes have been tested and not tx hash incongruences spotted. Approved ✅  and PR can be merged. Thanks!

PR #377, opened 2024-02-07
fix: blockHeight, absoluteSlot and blockTime should actually be withi…

matiwinnetou:
…n type: long

### DIFF
--- a/backend-modules/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosTransactionService.java
+++ b/backend-modules/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosTransactionService.java
@@ -8,7 +8,10 @@ import com.bloxbean.cardano.client.plutus.spec.RedeemerTag;
 import org.apache.commons.collections4.ListUtils;
 import rest.koios.client.backend.api.base.common.Asset;
 import rest.koios.client.backend.api.transactions.TransactionsService;
-import rest.koios.client.backend.api.transactions.model.*;
+import rest.koios.client.backend.api.transactions.model.TxCertificate;
+import rest.koios.client.backend.api.transactions.model.TxIO;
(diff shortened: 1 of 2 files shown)


backend-modules/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosTransactionService.java
```diff
@@ -91,9 +94,15 @@ private TransactionContent convertToTransactionContent(TxInfo txInfo) {
         TransactionContent transactionContent = new TransactionContent();
         transactionContent.setHash(txInfo.getTxHash());
         transactionContent.setBlock(txInfo.getBlockHash());
-        transactionContent.setBlockHeight(txInfo.getBlockHeight());
-        transactionContent.setBlockTime(txInfo.getTxTimestamp());
-        transactionContent.setSlot(txInfo.getAbsoluteSlot());
+        if (txInfo.getBlockHeight() != null) {
```
matiwinnetou, line 97: Can we make assumption that any of these value will never be null?  I find my own defensive programming here a bit clunky.

satran004: @matiwinnetou Thanks for the PR.

These should not be null unless there is a defect on the provider's side. But I think we can keep this null check.

satran004 (approved):

edridudi (approved):
Will Change data types in Koios Java Client as well for the next version:
https://github.com/cardano-community/koios-java-client/issues/161

backend-modules/koios/src/main/java/com/bloxbean/cardano/client/backend/koios/KoiosTransactionService.java
matiwinnetou: ok, feel free to merge when you think this is fine and right time for it.

PR #398, opened 2024-05-22
fix: Handle serialization and deserialization of bigint/bytestring when length > 64 

satran004:
**Issues:**

1. PlutusData.deserialize() doesn't handle BigInteger properly when it is encoded as a ByteString. Instead, it decodes `BigInteger` as `BytesPlutusData`.

BigInteger can also be represented as bytes.

```
big_int = int / big_uint / big_nint
big_uint = #6.2(bounded_bytes)
big_nint = #6.3(bounded_bytes)
```

2. All bytestring must be chunked up to 64 bytes (https://github.com/bloxbean/cardano-client-lib/issues/399)

**Fix**

- Read Tag of the ByteString to identify if it's a positive and negative number
- Custom encoder for chunked bytestring

**Example (Datum):**

```
d8798b581c217a7bf7a6494a0165ae2cb77aabeadb865d33d5be5f9f57b9b0e8f6581c276ba8bb7fe700d0d21b7da11edeb650729efe0b1595af0e643e31844b4e656f6e50756e6b3033311a3b9aca005563687447785a454339663038514d504b6970773073d87980d87980c2410040c241005820dee0b6f2b41b1a75244ec7b83c77c45d5cda5ad4ff3a4964340bb6089d5256e5
```


satran004:
Fyi, @nau

Removed `encodeAsByteString()`

Sotatek-HuyLe3a (approved):

satran004:
To fix: #399

Issue #429, opened 2024-09-10
AikenTransactionEvaluator slot conversion isssue (non-mainnet networks)


nemo83:
Currently AikenTransactionEvaluator, can't properly convert slot into timestamp for non-mainnet networks.

This prevent to correctly execute and validate contract which have interval validity checks on non-mainnet networks.

PR #431, opened 2024-09-10
Introduced SlotConfigs for Transaction Evaluators

nemo83:
Currently local transaction evaluators like AikentTransactionEvalutator and ScalusTE, only work for mainnet in case of contracts which require time interval validaiton.

This PR contribute to fix https://github.com/bloxbean/cardano-client-lib/issues/429


satran004 (approved):
Thanks @nemo83

PR #434, opened 2024-09-12
fix: Handle null or empty scriptHash in getScript method

### LINKED ISSUE #430: ScriptSupplier called w/ null
There seems to be a weird case where ScriptSupplier is called with script hash equals to null.

This seems to be due (just a guess!!) to the fact that, there is 1 contract involved, but its code it's attached via `attachSpendingValidator` but also a ref input is attached (this only contains a datum which represent protocols settings)

```
var tx = new ScriptTx()
                .collectFrom(utxo, ConstrPlutusData.of(1))
                .payToAddress(payeeAccount.baseAddress(), Amount.ada(5))
                .payToContract(userRecurringPaymentsAddress.getAddress(), ValueUtil.toAmountList(backToContract), datum)
                .readFrom(TransactionInput.builder()
                        .transactionId("df8707a513918eed71aec27f2fa83e63ef2410a59261ebfa02d2a5c3b022e3db")
                        .index(0)
                        .build())
                .withChangeAddress(adminAccount.baseAddress())
                .attachSpendingValidator(recurringPaymentsContract);
```

By manually wrapping teh script supplier and implementing a null check + returning empty, fixed the issue, so, I think, the caller it's just making an unnecessary call.

satran004:
To fix #430

### DIFF
--- a/backend/src/main/java/com/bloxbean/cardano/client/backend/api/DefaultScriptSupplier.java
+++ b/backend/src/main/java/com/bloxbean/cardano/client/backend/api/DefaultScriptSupplier.java
@@ -20,6 +20,9 @@ public class DefaultScriptSupplier implements ScriptSupplier {
 
     @Override
     public Optional<PlutusScript> getScript(String scriptHash) {
+        if (scriptHash == null || scriptHash.isEmpty())
+            return Optional.empty();
+
         try {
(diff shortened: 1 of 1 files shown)


nemo83 (approved):
🚀 

Thanks!

PR #451, opened 2024-09-30
Add support for Optional types in blueprint processing

### LINKED ISSUE #452: Constr alternative overwritten with 0 instead of proper value
It looks there is a tiny bug when added element to an object whose ConstrData alternative is not 0.

Screenshot 2024-09-30 at 15 15 41

satran004:
- Implemented changes to handle Optional types in blueprint definitions. 
- Fix #452 
- snake case to camel case


nemo83:
Things are looking much better, but not perfect yet.

Can you please try to generate blueprint for this contract? https://github.com/easy1staking-com/cardano-recurring-payment/blob/main/plutus.json

It's a v3 multi contract project.

Some issues:
* snake case is still appearing, in the `Address` class for instance we have an attribute `private PaymentCredential payment_credential;`
* always in class `Address` the import of `PaymentCredential` is missing.

There could be more, but I'm blocked here. We're so so close!

Sotatek-HuyLe3a (approved):

PR #454, opened 2024-10-09
Refactor COSESign builders for external payload and hashing

satran004:
- Refactor `COSESignBuilder` and `COSESign1Builder` to handle external payloads and add support for payload hashing. 
- Update `CIP30DataSigner` logic to remove redundant hashed payload header. 
- Adjust tests to align with these changes and add new test files for Rust code comparisons.

### DIFF
--- a/cip/cip8/src/main/java/com/bloxbean/cardano/client/cip/cip8/builder/COSESign1Builder.java
+++ b/cip/cip8/src/main/java/com/bloxbean/cardano/client/cip/cip8/builder/COSESign1Builder.java
@@ -26,12 +26,19 @@ public class COSESign1Builder {
 
     public SigStructure makeDataToSign() {
         Headers headersCopy = headers.copy();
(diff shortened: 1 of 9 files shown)


cip/cip8/src/main/java/com/bloxbean/cardano/client/cip/cip8/builder/COSESign1Builder.java
```diff
@@ -26,12 +26,19 @@ public COSESign1Builder(Headers headers, byte[] payload, boolean isPayloadExtern
 
     public SigStructure makeDataToSign() {
         Headers headersCopy = headers.copy();
+        headersCopy.unprotected().addOtherHeader("hashed", hashed ? SimpleValue.TRUE : SimpleValue.FALSE);
```
matiwinnetou, line 29: maybe we only set TRUE to avoid extra storage?

matiwinnetou (approved):

PR #455, opened 2024-10-10
Adjust the fee calculation when reference script bytes are present in the input UTXOs of a transaction

### LINKED ISSUE #450: Adjust the fee calculation when reference script bytes are present in the input UTXOs of a transaction.
In the Conway era, if an input with a script reference is used as a reference input in a transaction, the fee calculation logic considers the reference script size and uses `minFeeRefScriptCostPerByte` to calculate the final fee.

However, there could be a scenario where someone tries to spend an input with script reference bytes. It seems that the same logic may need to be applied during fee calculation.

So, if there is `FeeTooSmallUTxO` error, please check if you are trying to spend an input with script reference bytes.

satran004:
To fix #450 

1. A script reference can be part of the regular inputs in a script transaction. **For example**, if the reference script is part of the spending UTXO, a reference input is not required.

2. Update the `FeeCalculator` to calculate the reference script size in inputs and adjust the fee accordingly.

3. If both a script reference and a script witness (`transaction.getWitnessSet().getPlutusXScript()`) are present, `ExtraneousScriptWitnessesUTXOW` is thrown. To fix this, `DuplicateScriptWitnessChecker` is introduced, which returns a composable function to remove any duplicate witnesses. 

4. A new API `removeDuplicateScriptWitnesses(boolean flag)` has been added to `QuickTxBuilder` to enable the removal of duplicate witnesses in the above scenario. By default, this flag is false, but it can be enabled by calling the above method if required.

```java
Result<String> result1 = quickTxBuilder.compose(scriptTx)
    ...
    .removeDuplicateScriptWitnesses(true)
    .completeAndWait(System.out::println);
```
5. A new overloaded method in `Tx` api.

```
public T payToContract(String address, Amount amount, PlutusData datum, Script refScript)
```

### DIFF
--- a/common-spec/src/main/java/com/bloxbean/cardano/client/spec/Script.java
+++ b/common-spec/src/main/java/com/bloxbean/cardano/client/spec/Script.java
@@ -1,12 +1,6 @@
 package com.bloxbean.cardano.client.spec;
 
-import co.nstant.in.cbor.CborException;
-import co.nstant.in.cbor.model.Array;
-import co.nstant.in.cbor.model.ByteString;
 import co.nstant.in.cbor.model.DataItem;
-import co.nstant.in.cbor.model.UnsignedInteger;
-import com.bloxbean.cardano.client.common.cbor.CborSerializationUtil;
-import com.bloxbean.cardano.client.exception.CborRuntimeException;
 import com.bloxbean.cardano.client.exception.CborSerializationException;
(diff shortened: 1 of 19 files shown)


nemo83 (approved):
🔥

fabianbormann (approved):

Sotatek-HuyLe3a (approved):

PR #464, opened 2024-10-16
feat: support for CIP-30 payload hash with an extra overloaded methods

satran004:
Duplicate of https://github.com/bloxbean/cardano-client-lib/pull/453


nemo83 (approved):
approving although I'm less familiar with this PR

PR #499, opened 2025-04-08
fix: #498 Use dummy key and signature for additional signers instead of generating new dummy accounts

### LINKED ISSUE #498: `.additionalSignersCount` performance
`.additionalSignersCount` method under the hood spawns HD wallets, and it can eat up to 70% of CPU time crunching SHA512 hashes: https://github.com/bloxbean/cardano-client-lib/blob/6f0df0678e4ad20c04af8bf74d016ecb9509af00/function/src/main/java/com/bloxbean/cardano/client/function/helper/FeeCalculators.java#L197

This is needed for calculating size only, so there might be several ways to go:
* Do the calculation directly (obviously the best one, but probably too intrusive).
* Use static fake signatures.
* Introduce a cache of dummy accounts, not to create them each time.

satran004:
To fix #498 

- Use dummy key and signature bytes for additional signers instead of generating new dummy accounts.
- Update the last two bytes of the dummy key and signature to ensure uniqueness.

### DIFF
--- a/function/src/main/java/com/bloxbean/cardano/client/function/helper/FeeCalculators.java
+++ b/function/src/main/java/com/bloxbean/cardano/client/function/helper/FeeCalculators.java
@@ -1,11 +1,9 @@
 package com.bloxbean.cardano.client.function.helper;
 
-import com.bloxbean.cardano.client.account.Account;
 import com.bloxbean.cardano.client.api.exception.ApiException;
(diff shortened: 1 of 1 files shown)


euonymos:
Excellent, thank you, @satran004, I will bump the version soon.

Sotatek-HuyLe3a (approved):

PR #500, opened 2025-05-01
[FIX] Fixed Value subtractions with extra assets or multi assets in second operand

nemo83:
If second operand contains:
1. an additional asset within a multi asset which is not present in the first operand
2. and additional multi asset which is not present in the first operand

The subtraction skips them, instead of setting them as negative in the resulting value.

### DIFF
--- a/transaction-spec/src/test/java/com/bloxbean/cardano/client/transaction/spec/MultiAssetSpecTest.java
+++ b/transaction-spec/src/test/java/com/bloxbean/cardano/client/transaction/spec/MultiAssetSpecTest.java
@@ -66,7 +66,7 @@ class MultiAssetSpecTest {
     }
 
     @Test
-    void addSamePolicyPartialAssetOveral() {
+    void addSamePolicyPartialAssetOverlap() {
 
         Asset asset1 = Asset.builder().name("asset1").value(BigInteger.valueOf(100L)).build();
         Asset asset2 = Asset.builder().name("asset2").value(BigInteger.valueOf(200L)).build();
--- a/transaction-spec/src/test/java/com/bloxbean/cardano/client/transaction/spec/MultiAssetSpecTest.java
+++ b/transaction-spec/src/test/java/com/bloxbean/cardano/client/transaction/spec/MultiAssetSpecTest.java
@@ -49,6 +49,40 @@ class MultiAssetSpecTest {
     }
 
     @Test
+    void addSamePolicyDifferentAsset() {
+
+        Asset asset1 = Asset.builder().name("asset1").value(BigInteger.valueOf(100L)).build();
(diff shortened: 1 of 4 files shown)


transaction-spec/src/test/java/com/bloxbean/cardano/client/transaction/spec/MultiAssetSpecTest.java
```diff
@@ -177,4 +211,74 @@ void subtractMultiAssetListsWhenAdditionalMultiAssetInFirstList() {
 
         assertThat(result, equalTo(exList));
     }
+
+    @Test
+    void subtractMultiAssetListsWhenAdditionalAssetInSecondList() {
```
satran004, line 216: @nemo83 Shouldn't this be called `AdditionalPolicyInSecondList`, since there's an extra policy_id3 in the second list?

transaction-spec/src/test/java/com/bloxbean/cardano/client/transaction/spec/MultiAssetSpecTest.java
```diff
@@ -177,4 +211,74 @@ void subtractMultiAssetListsWhenAdditionalMultiAssetInFirstList() {
...
+        MultiAsset exMultiAsset2 = MultiAsset.builder().policyId("policy_id2").assets(Arrays.asList(exAsset3)).build();
+        MultiAsset exMultiAsset3 = MultiAsset.builder().policyId("policy_id3").assets(Arrays.asList(exAsset4)).build();
+
+        List<MultiAsset> exList = Arrays.asList(exMultiAsset1, exMultiAsset2, exMultiAsset3);
+
+        assertThat(result, equalTo(exList));
+    }
+
+    @Test
+    void subtractMultiAssetListsWhenAdditionalPolicyInSecondList() {
```
satran004, line 252: In this case, there is an extra asset, `asset4`. Shouldn't the method be named `AdditionalAssetInSecondList`?

satran004:
@nemo83 Thanks. The changes look good to me. Just a few minor comments for typos—please check.
I’ve fixed two of them.

transaction-spec/src/test/java/com/bloxbean/cardano/client/transaction/spec/MultiAssetSpecTest.java
nemo83: yes correct

transaction-spec/src/test/java/com/bloxbean/cardano/client/transaction/spec/MultiAssetSpecTest.java
nemo83: yes correct

nemo83:
good catch, I planned the tests, and then mixed the name 😂

satran004 (approved):

PR #567, opened 2026-02-05
refactor:  implements a language-agnostic naming strategy to handle many different versions of CIP-57 blueprints (for various versions of Aiken compiler)

matiwinnetou:


satran004 (approved):
LGTM. Thanks

Issue #581, opened 2026-02-22
Unified Tx API: Merge ScriptTx operations into Tx


satran004:
## Summary

 Merge all `ScriptTx` operations into `Tx` to provide a single unified transaction builder API. `ScriptTx` will be deprecated for backward compatibility.

Original issue https://github.com/bloxbean/cardano-client-lib/issues/576

 ## Problem

 The `quicktx` module has two transaction builder classes. `Tx` for regular transactions and `ScriptTx` for script-based transactions. This split creates confusion for users:

 - Operations like registering a script-derived stake address don't need `ScriptTx` (any account can do it), but users expect to find them there
 - Users must know upfront whether their transaction involves scripts to pick the right class

 ### Before (current)
 ```java
 // User unsure: "Do I use Tx or ScriptTx to register a script stake address?"

 // Registration — use Tx (no script witness needed)
 Tx tx = new Tx()
     .registerStakeAddress(scriptStakeAddress)
     .from(senderAddress);

 // Deregistration — must use ScriptTx (needs script witness)
 ScriptTx stx = new ScriptTx()
     .deregisterStakeAddress(scriptStakeAddress, redeemer)
     .attachCertificateValidator(plutusScript);
```
### After (unified)

```
 // Everything in one class — no confusion

 Tx tx = new Tx()
     .registerStakeAddress(scriptStakeAddress)
     .from(senderAddress);

 Tx tx = new Tx()
     .deregisterStakeAddress(scriptStakeAddress, redeemer)
     .attachCertificateValidator(plutusScript)
     .from(senderAddress);

 // Mix script and non-script operations in a single Tx — previously impossible
 // Script ops (collectFrom with redeemer, Plutus mint) + non-script ops (payment, stake registration)
 // Deposit for stake registration comes from the fee payer address
 Tx tx = new Tx()
     .collectFrom(scriptUtxo, spendRedeemer)
     .attachSpendingValidator(spendingScript)
     .mintAsset(mintingScript, asset, mintRedeemer, receiverAddr)
     .payToAddress(receiverAddr, Amount.ada(10))
     .registerStakeAddress(someAddress)
     .from(feePayerAddress);
```

### Scope

Both branches need this change. Two separate PRs:

PR 1: master branch (intent-based architecture)
PR 2: release/0.7.x branch (monolithic architecture)

Issue #576, opened 2026-02-16
ScriptTx lacks registerStakeAddress methods


nemo83:
As per title, ScriptTx has the unregister but lacks the `registerStakeAddress` methods.

nemo83:
Why do we have Tx vs ScriptTx in first place? Is it a dev-ex thing or does it have technical implications?

If I read your message correctly, there should be no technical requirements, in fact I added https://github.com/bloxbean/cardano-client-lib/compare/v0.7.1...feat/add-register-stake-address?expand=1 and it just worked.

If we really wanna keep a Tx vs ScripTx, ScriptTx should be allowed to do everything Tx does PLUS Script specific tx, not less.

To me a transaction is a transaction, and the fact that I need to combine 2 tx is problematic because I completely lose control of outputs, that many times need to be ordered in a particular way. I think this is also not the case with Mesh, so in the effort of providing a consistent UX across libraries maybe it would be worth to consolidate the tx building experience?

satran004:
The original idea was to keep ScriptTx separate, since some attributes are not valid for ScriptTx (or vice versa). For example, `from` is specific to Tx.
We can always revisit this. Btw, the current design follows composition, so we can introduce another composite layer that includes both if needed.

PR #582, opened 2026-02-23
feat: Unify Tx and ScriptTx into a single Tx class

satran004:
To fix: #581 #576 

 ## Summary

  - Merge all `ScriptTx` script operations (`collectFrom`, `mintAsset`, reference inputs, UTXO filter strategies) into the `Tx` class, allowing both regular and script-based operations in a single transaction builder
  - Deprecate `ScriptTx` — all existing `ScriptTx` usage continues to work but new code should use `Tx` exclusively
  - Add `hasRedeemer()` to the `TxIntent` interface with empty-string guard for `redeemerHex` across all 11 intent implementations

 ## Motivation

  Previously, users had to choose between `Tx` (for simple transfers, staking, governance) and `ScriptTx` (for Plutus script interactions) at construction time. This made it impossible to combine regular and script operations in a single `Tx` instance.

  The unified `Tx` now handles script detection automatically via `hasScriptIntents()`, which triggers Plutus infrastructure (collateral, ex-unit evaluation, redeemer index adjustment) only when needed.

  ## Key changes

  **`Tx.java`** — Added all script related methods:
  - `collectFrom()` variants (single/list UTXOs, predicate-based, filter strategies)
  - `mintAsset()` / `mintAssets()` variants
  - `referenceInput()`, `readFrom()` for reference inputs
  - Script attachment methods (`attachSpendingValidator`, `attachMintingPolicy`, etc.)

  **`AbstractTx.java`** — Moved shared script logic up from `ScriptTx`:
  - `hasScriptIntents()` detection method
  - `verifyAndAdjustRedeemerIndexes()` in `preTxEvaluation` / `postBalanceTx`
  - `setDefaultFrom()` hook for fee-payer address propagation

  **`QuickTxBuilder.java`** — Generalized script handling:
  - Changed `instanceof ScriptTx` checks to `tx instanceof ScriptTx || tx.hasScriptIntents()`
  - Uses `setDefaultFrom()` instead of casting to `ScriptTx`

  **`TxIntent.java`** + 11 intent classes — Added `hasRedeemer()`:
  - Default returns `false`; overridden in each intent that supports redeemers
  - Guards against empty `redeemerHex` strings from YAML deserialization

  **`ScriptTx.java`** — Marked `@Deprecated`, removed duplicated logic
  
  
## Deposit Resolution Changes
  
- Centralized deposit resolution: Introduces DepositMode and DepositResolvers to handle protocol deposits (stake/pool/DRep registration, governance proposals) in a dedicated Phase 4 of AbstractTx.complete(), replacing duplicated deposit logic scattered across individual intent classes.
- Configurable deposit payer: Adds TxContext.depositPayer(address) and TxContext.depositMode(mode) to control who pays deposits and how funds are sourced (AUTO, CHANGE_OUTPUT, ANY_OUTPUT, NEW_UTXO_SELECTION).
- ChangeOutput marker: Adds ChangeOutput subclass in transaction-spec so deposit resolution can distinguish change outputs from user-declared outputs.

### DIFF
--- a/quicktx/adr/unified-tx-and-deposit-resolution.md
+++ b/quicktx/adr/unified-tx-and-deposit-resolution.md
@@ -1,0 +1,529 @@
+# Unified Tx API & Deposit Resolution
+
+**Status**: Implemented
+**Modules**: `quicktx`, `transaction-spec`
+**Since**: Cardano Client Library 0.8.0-preview
+
+## 1. Overview
+
+This document describes two closely related changes to the QuickTx transaction-building API:
+
+1. **Unified Tx API** — All smart-contract operations (`collectFrom`, `mintAsset`, `attachValidator`, etc.) are now available directly on the `Tx` class. The separate `ScriptTx` class is deprecated. Users no longer need to decide upfront whether a transaction involves scripts.
+
+2. **Unified Deposit Resolution** — Protocol deposits (stake registration, pool registration, DRep registration, governance proposals) are resolved through a single Phase 4 step that works identically for all transaction types. There are no dummy outputs and no separate code paths for script vs. non-script transactions.
+
+Together, these changes mean you can build any Cardano transaction — from a simple ADA transfer to a multi-script governance action with deposits — using a single `Tx` class and a predictable four-phase build pipeline.
+
+---
+
+## 2. Unified Tx API
+
+### 2.1 What Changed
+
+Previously, building a transaction that involved Plutus scripts required using `ScriptTx`:
+
+```java
+// OLD — had to know upfront that you need ScriptTx
+ScriptTx scriptTx = new ScriptTx()
+    .collectFrom(utxo, redeemer)
+    .payToAddress(receiver, Amount.ada(10))
+    .attachSpendingValidator(script)
+    .withChangeAddress(changeAddr);
+```
+
+If you started with `Tx` and later realised you needed a script input, you had to rewrite using `ScriptTx`. This created confusion and made it impossible to mix script and non-script operations in a single transaction object.
+
+Now, all script operations live directly on `Tx`:
+
+```java
+// NEW — one class for everything
+Tx tx = new Tx()
+    .collectFrom(utxo, redeemer)           // script input
+    .payToAddress(receiver, Amount.ada(10)) // regular payment
+    .attachSpendingValidator(script)
+    .withChangeAddress(changeAddr);
+```
+
+### 2.2 Before and After
+
+**Before** — Separate classes forced early decisions:
+```java
+// Simple payment → Tx
+Tx simpleTx = new Tx()
+    .payToAddress(receiver, Amount.ada(50))
+    .from(sender);
+
+// Script interaction → ScriptTx (different class!)
+ScriptTx scriptTx = new ScriptTx()
+    .collectFrom(scriptUtxo, redeemer)
+    .payToAddress(receiver, Amount.ada(10)) 
+    .attachSpendingValidator(validator);
+
+// Could NOT mix in one object
+```
+
+**After** — Everything in one class:
+```java
+// Simple payment
+Tx simpleTx = new Tx()
+    .payToAddress(receiver, Amount.ada(50))
+    .from(sender);
+
+// Script interaction — same class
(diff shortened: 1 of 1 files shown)


quicktx/src/main/java/com/bloxbean/cardano/client/quicktx/intent/DRepDeregistrationIntent.java
```diff
@@ -287,4 +287,9 @@ public TxBuilder apply(IntentContext ic) {
             }
         };
     }
+
+    @Override
+    public boolean hasRedeemer() {
+        return redeemer != null || (redeemerHex != null && !redeemerHex.isEmpty());
+    }
```
nemo83: duplicated code. could be worth pushing up or creating a helper function for reusability.

quicktx/src/main/java/com/bloxbean/cardano/client/quicktx/AbstractTx.java
```diff
@@ -465,19 +504,100 @@ private TxBuilder buildInputBuildersFromUtxoSupplier(TxOutputBuilder txOutputBui
...
+     * Includes fix for P0 bug: filters for present Optionals before findFirst().
+     *
+     * @param redeemer the redeemer to find the UTXO for
+     * @return the matching UTXO if found
+     */
+    protected Optional<Utxo> getUtxoForRedeemer(Redeemer redeemer) {
+        if (intentions == null) return Optional.empty();
+        return intentions.stream()
+                .filter(intention -> intention instanceof ScriptCollectFromIntent)
+                .map(intention -> ((ScriptCollectFromIntent) intention).getUtxoForRedeemer(redeemer))
```
nemo83: flatMap and you can avoid double case:

if true return Stream.of()
else Stream.empty

quicktx/src/main/java/com/bloxbean/cardano/client/quicktx/intent/DRepDeregistrationIntent.java
satran004: @nemo83 
Yes, I have thought about it. Most intents have redeemer and redeemerHex, but there are a few intents that also have an additional redeemer variant. We will clean this up in a future PR.

quicktx/src/main/java/com/bloxbean/cardano/client/quicktx/AbstractTx.java
```diff
@@ -358,6 +375,23 @@ boolean hasMultiAssetMinting() {
...
+     * and any intent that has a redeemer set (e.g., staking/governance intents with script witnesses).
+     *
+     * @return true if the transaction has script intents
+     */
+    public boolean hasScriptIntents() {
+        if (intentions == null || intentions.isEmpty()) return false;
+        return intentions.stream().anyMatch(intent ->
+                intent instanceof ScriptCollectFromIntent ||
+                intent instanceof ScriptMintingIntent ||
+                intent.hasRedeemer()
```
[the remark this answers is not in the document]
satran004: It should. In the initial implementation we were using instance based check. We will verify again and remove that.

quicktx/src/main/java/com/bloxbean/cardano/client/quicktx/AbstractTx.java
```diff
@@ -465,19 +504,100 @@ private TxBuilder buildInputBuildersFromUtxoSupplier(TxOutputBuilder txOutputBui
...
+     * Verify and adjust redeemer indexes in the transaction to match the sorted input order.
+     * Moved from ScriptTx to support unified Tx with script intents.
+     *
+     * @param transaction the transaction to adjust
+     */
+    protected void verifyAndAdjustRedeemerIndexes(Transaction transaction) {
+        if (transaction.getWitnessSet().getRedeemers() == null) return;
+        for (Redeemer redeemer : transaction.getWitnessSet().getRedeemers()) {
+            if (redeemer.getTag() != RedeemerTag.Spend)
+                continue;
```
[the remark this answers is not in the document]
satran004: It seems this fix is missing in master branch :( https://github.com/bloxbean/cardano-client-lib/pull/541

PR #585, opened 2026-02-25
fix: Use Tuple types (Pair through Quintet) for multi-item tuples and fix element-level PlutusData handling

matiwinnetou:
**Summary**
  - Multi-item tuples previously resolved to `List`, causing nonexistent `java.lang.converter.ObjectConverter` references. Root causes: `$ref` resolution via `copyFrom()`
  overwrites "Tuple" title with field name (breaking title-based detection), and `ListDataTypeProcessor` bypassed `resolveType()` by calling `resolveListType()` directly
  - Tuple resolution now maps by size: 2-item → `Pair`, 3-item → `Triple<T1,T2,T3>`, 4-item → `Quartet<T1,T2,T3,T4>`, 5-item → `Quintet<T1,T2,T3,T4,T5>`; 6+ items throw
  `BlueprintGenerationException`
  - Activated existing `Triple.java`, `Quartet.java`, `Quintet.java`: made public with full ser/deser codegen support
  - Fixed variable collision: hardcoded `var data = constr.getData()` collided with fields named "data" (from `#/definitions/Data` references) — renamed to `var constrData`
  - Added `PLUTUSDATA` case in `toPlutusDataCodeBlock()` and `fromPlutusDataToObj()` for raw pass-through (previously caused converter lookup for nonexistent class)
  - Auto-generate converters for shared/registered blueprint types
  - Added `toPlutusData()` to `ByteArrayWrapper` and delegated in BYTES converter
  - Added `integrationTestNoRegistry` Gradle task for testing without shared type registry

### DIFF
--- a/annotation-processor/src/main/java/com/bloxbean/cardano/client/plutus/annotation/processor/ClassDefinitionGenerator.java
+++ b/annotation-processor/src/main/java/com/bloxbean/cardano/client/plutus/annotation/processor/ClassDefinitionGenerator.java
@@ -7,6 +7,8 @@ import com.bloxbean.cardano.client.plutus.annotation.processor.exception.NotSupp
 import com.bloxbean.cardano.client.plutus.annotation.processor.model.*;
 import com.bloxbean.cardano.client.plutus.spec.PlutusData;
 import com.bloxbean.cardano.client.plutus.blueprint.type.Pair;
+import com.bloxbean.cardano.client.plutus.blueprint.type.Quartet;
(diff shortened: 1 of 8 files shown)


satran004:
@matiwinnetou The changes look good to me. You need to resolve conflict to merge.

Do we also need to support quadruple and quintuple, since we already have those classes in types package?
If the number of items is more than that, maybe we can generate a generic class with field1...fieldn instead of relying on shared types.

PR #619, opened 2026-03-27
 Fix MPF RocksDB audit issues: performance, resource safety, and observability

satran004:
## Summary

  Addresses findings from a detailed code review.

  - **Staged cache key encoding** — Replace `Arrays.toString(hash)` with `HexUtil.encodeHexString()`/`Bytes.toHex()` in `RocksDbNodeStore`, `RocksDbGc`, and test `TestNodeStore`. 

  - **GC resource leak** — Move `WriteOptions` and `WriteBatch` into try-with-resources in `RocksDbGc.sweepUnreachable()` to prevent native memory leaks on exception paths.
  - **withBatch() nesting guard** — Add `IllegalStateException` guard against nested `withBatch()` calls on the same instance, which would silently corrupt the outer batch's ThreadLocal state. Replace fragile save/restore of `TL_STAGED` with clean `remove()`.
  Verified all 8 callers — none nest on the same object.
  - **Missing node observability** — Add `log.warn()` in `PutOperationVisitor` and `DeleteOperationVisitor` (5 locations) when a referenced node hash is missing from storage. Previously this was silently handled, making storage corruption undetectable.

### DIFF
--- a/verified-structures/merkle-patricia-forestry-rocksdb/src/main/java/com/bloxbean/cardano/vds/mpf/rocksdb/gc/internal/RocksDbGc.java
+++ b/verified-structures/merkle-patricia-forestry-rocksdb/src/main/java/com/bloxbean/cardano/vds/mpf/rocksdb/gc/internal/RocksDbGc.java
@@ -1,5 +1,6 @@
 package com.bloxbean.cardano.vds.mpf.rocksdb.gc.internal;
 
+import com.bloxbean.cardano.client.util.HexUtil;
 import com.bloxbean.cardano.vds.mpf.internal.NodeRefParser;
 import com.bloxbean.cardano.vds.mpf.rocksdb.RocksDbNodeStore;
 import com.bloxbean.cardano.vds.mpf.rocksdb.RocksDbRootsIndex;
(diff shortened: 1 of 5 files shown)


verified-structures/merkle-patricia-forestry-rocksdb/src/main/java/com/bloxbean/cardano/vds/mpf/rocksdb/gc/internal/RocksDbGc.java
```diff
@@ -48,33 +49,36 @@ public static Set<String> markReachable(RocksDbNodeStore store, Collection<byte[
      */
     public static Report sweepUnreachable(RocksDbNodeStore store, Set<String> reachable, Options opts) {
         Report report = new Report();
-        try (ReadOptions ro = new ReadOptions(); RocksIterator it = store.db().newIterator(store.nodesHandle(), ro)) {
-            WriteOptions wopts = new WriteOptions();
+        try (ReadOptions ro = new ReadOptions();
+             RocksIterator it = store.db().newIterator(store.nodesHandle(), ro);
```
Sotatek-HuyLe3a, line 53: You might want to adjust the code formatting in this section, it looks a bit over-indented. line 53-54

Sotatek-HuyLe3a (approved):

PR #624, opened 2026-04-27
 fix: prefer raw cost models for protocol params

### LINKED ISSUE #622: BFBackendService reads Blockfrost's legacy cost_models — truncated/reordered on Conway, causes ScriptIntegrityHashMismatch
## Summary

`ProtocolParams` deserializes Blockfrost's `/epochs/latest/parameters` via the `cost_models` named view (`LinkedHashMap<String, LinkedHashMap<String, Long>> costModels`). Post-Chang/Plomin, Blockfrost's named `cost_models` is:

- **Truncated**: missing ~53 Plutus V3 parameters added in Plomin (297 named keys vs 350 canonical integers)
- **Reordered**: alphabetical ≠ canonical Plutus-builtin index order

`CostModelUtil.getCostModelFromProtocolParams` then produces a `CostModel` that doesn't match what the ledger uses. Transactions with Plutus scripts are rejected with `ScriptIntegrityHashMismatch`.

## Affected

- `cardano-client-lib` 0.7.x / 0.8.0-preN using `BFBackendService` as `ProtocolParamsSupplier`
- All networks post-Chang (preview / preprod / mainnet)
- Any `QuickTxBuilder` / transaction flow that uses Plutus scripts
- Verified on preview at epoch 1274; BF output byte-compared against Ogmios `queryLedgerState/protocolParameters`

## Repro

```bash
curl -s -H 'project_id: <preview_project_id>' \
  https://cardano-preview.blockfrost.io/api/v0/epochs/latest/parameters | \
  jq '{v3_named: (.cost_models.PlutusV3 | length),
       v3_raw:   (.cost_models_raw.PlutusV3 | length)}'
# => { "v3_named": 297, "v3_raw": 350 }
```

`cost_models_raw` is Blockfrost's canonical flat-array view and matches Ogmios byte-for-byte. `cost_models` has the Conway gap above.

## Root cause

- `core-api/src/main/java/com/bloxbean/cardano/client/api/model/ProtocolParams.java:46` — only `costModels` field (Jackson binds to BF's `cost_models` by default snake_case→camelCase); no `costModelsRaw`.
- `core-api/src/main/java/com/bloxbean/cardano/client/api/util/CostModelUtil.java:642-677` — `getCostModelFromProtocolParams` reads `protocolParams.getCostModels().get(\"PlutusV3\")`. The heuristic `sortByAsIntegerKey = costModelMap.containsKey(\"0\") && costModelMap.containsKey(\"1\")` is **false** for Blockfrost's named view (keys look like `addInteger-cpu-arguments-intercept`), so the code falls through to `LinkedHashMap` insertion order = alphabetical = wrong.

Koios backend (`KoiosEpochService`) appears unaffected — it handles arrays differently.

## Observed failure (real preview tx)

```
providedScriptIntegrity: 3702f2f2fad49170706bc1f04ed46b56e4111edd81d044f028da3987a3c75396
computedScriptIntegrity: 32169542440774d7f77416b62c25a887bf440e81d74e79fbc8803b50443cd8b7
```

The `provided` hash reproduces from the 297-entry alphabetical V3 cost model; the `computed` hash reproduces from the 350-entry canonical V3 cost model.

## Suggested fix

1. Add the canonical field to the DTO (`ProtocolParams.java`):

```java
@JsonProperty(\"cost_models_raw\")
private LinkedHashMap<String, List<Long>> costModelsRaw;
```

2. Prefer it in `CostModelUtil.getCostModelFromProtocolParams`:

```java
if (protocolParams.getCostModelsRaw() != null) {
    List<Long> raw = protocolParams.getCostModelsRaw().get(languageKey);
    if (raw != null) {
        long[] costModel = raw.stream().mapToLong(Long::longValue).toArray();
        return Optional.of(new CostModel(language, costModel));
    }
}
// ... existing named-map fallback for older BF deployments
```

This brings V3 from 297 wrong-order entries to 350 canonical entries. Equivalent fix verified on `@evolution-sdk/evolution` against the same failing preview tx — tx submits successfully after the fix.

Cross-reference: https://github.com/IntersectMBO/evolution-sdk/issues/303

Happy to open a PR.

satran004:
To fix #622 
 
 ## Summary

  Fix protocol parameter cost model handling to prefer canonical raw cost model arrays when available.

  Blockfrost’s named `cost_models` can be truncated/reordered for Conway/Plomin cost models, which can produce an incorrect script integrity hash. This change adds `costModelsRaw` support and updates cost model extraction to use raw arrays first.

  ## Changes

  - Add `costModelsRaw` to `ProtocolParams` for `cost_models_raw`
  - Prefer `costModelsRaw` in `CostModelUtil`
  - Keep existing `costModels` fallback for Blockfrost-compatible APIs that do not expose raw cost models yet, such as current Yaci Store/Yaci DevKit
  - Populate raw cost models in Koios and Ogmios providers
  - Add tests covering raw preference and named fallback behavior

### DIFF
--- a/core-api/src/main/java/com/bloxbean/cardano/client/api/util/CostModelUtil.java
+++ b/core-api/src/main/java/com/bloxbean/cardano/client/api/util/CostModelUtil.java
@@ -7,6 +7,7 @@ import com.bloxbean.cardano.client.plutus.spec.Language;
 
 import java.util.Comparator;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Stream;
@@ -647,6 +648,17 @@ public class CostModelUtil {
             languageKey = "PlutusV2";
         } else if (language == Language.PLUTUS_V3) {
             languageKey = "PlutusV3";
+        }
+
+        if (protocolParams.getCostModelsRaw() != null) {
+            List<Long> raw = protocolParams.getCostModelsRaw().get(languageKey);
+            if (raw != null && !raw.isEmpty()) {
+                long[] costModel = raw.stream()
+                        .mapToLong(Long::longValue)
+                        .toArray();
+
+                return Optional.of(new CostModel(language, costModel));
+            }
         }
 
         if (protocolParams.getCostModels() == null)
--- a/backend-modules/koios/src/it/java/com/bloxbean/cardano/client/backend/koios/it/KoiosEpochServiceIT.java
+++ b/backend-modules/koios/src/it/java/com/bloxbean/cardano/client/backend/koios/it/KoiosEpochServiceIT.java
@@ -106,8 +106,8 @@ class KoiosEpochServiceIT extends KoiosBaseTest {
         assertThat(protocolParams.getDrepActivity(), greaterThan(0));
         assertThat(protocolParams.getMinFeeRefScriptCostPerByte(), greaterThan(BigDecimal.ZERO));
 
-        assertThat(protocolParams.getCostModels().get("PlutusV1").size(), is(166));
-        assertThat(protocolParams.getCostModels().get("PlutusV2").size(), is(175));
-        assertThat(protocolParams.getCostModels().get("PlutusV3").size(), greaterThan(251));
+        assertThat(protocolParams.getCostModels().get("PlutusV1").size(), greaterThanOrEqualTo(166));
+        assertThat(protocolParams.getCostModels().get("PlutusV2").size(), greaterThanOrEqualTo(175));
+        assertThat(protocolParams.getCostModels().get("PlutusV3").size(), greaterThanOrEqualTo(251));
     }
 }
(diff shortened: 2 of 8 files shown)


core-api/src/main/java/com/bloxbean/cardano/client/api/util/CostModelUtil.java
```diff
@@ -649,6 +650,17 @@ public static Optional<CostModel> getCostModelFromProtocolParams(ProtocolParams
...
+            if (raw != null && !raw.isEmpty()) {
+                long[] costModel = raw.stream()
+                        .mapToLong(Long::longValue)
+                        .toArray();
+
+                return Optional.of(new CostModel(language, costModel));
+            }
+        }
+
         if (protocolParams.getCostModels() == null)
```
nemo83, line 664: should this be femoved?

nemo83 (approved):
just one question?

core-api/src/main/java/com/bloxbean/cardano/client/api/util/CostModelUtil.java
satran004: We will keep this for now to minimize changes. But, since costModels is now deprecated in CCL, we will remove the entire costModels section later, once all dependent components (e.g., Yaci Store) have migrated to costModelRaw.

PR #629, opened 2026-05-28
Modernise Plutus annotation processor: latest Aiken syntax and stdlib >= v3 only + devnet tests

matiwinnetou:
## Summary

Modernises the Plutus annotation processor to target **Aiken stdlib v3.x only**, then proves the result works end-to-end with a battery of devnet tests against a real local Cardano network. The branch has three phases:

1. **Drop legacy support** — remove parallel code paths for Aiken stdlib v1 / v2 and the `$`-syntax generic delimiter that came with them.
2. **Verify v3.1 alignment** — confirm the v3 registry covers stdlib 3.1 unchanged (3.1 added only functions, no new shared types).
3. **Prove the codegen works on a real ledger** — add 11 devnet tests that cover every distinct codegen surface, all running against Yaci DevKit in CI.

14 commits, all green on CI.

## ⚠️ Breaking change for users on stdlib v1 / v2

Blueprints compiled with Aiken `< 1.1` (stdlib v1 / v2 era) **stop resolving** through the registry after this PR. The fix on the user side: re-compile the contract with a modern Aiken (`>= 1.1.21`, stdlib `>= 3.0`). The blueprint JSON itself is the only thing that changes — Java code referencing generated classes does not.

The processor will not silently miscompile a legacy blueprint; if you point a legacy blueprint at the new processor you'll get explicit codegen failures, not subtly wrong CBOR.

### Phase 1 — drop legacy Aiken support

**`plutus-aiken`**

- Deleted: `OutputReferenceV1`, `ReferencedCredential`, V1-shape `Credential`
- Rewrote `Address` from V1 (`Credential` + `Optional<ReferencedCredential>`) to V3 (`PaymentCredential` + `Optional<StakeCredential>`)
- `AikenBlueprintTypeRegistry`: single-map dispatch, empty `annotationHints()`, V3 schema builders renamed without the `V3` suffix
- Kept `@AikenStdlib` annotation + `AikenStdlibVersion` enum as a single-value (`V3`) source-compat marker — no-op for the processor today, kept as an extension point for a future stdlib v4+
- `AikenBlueprintTypeRegistryTest` rewritten for V3-only

**`annotation-processor`**

- Stripped `$`-syntax detection from `BlueprintAnnotationProcessor`, `FieldSpecProcessor`, `BlueprintUtil`, `JavaFileUtil`, `DefaultNamingStrategy`
- Removed `@AikenStdlib(...)` usage from the 9 kept blueprint interfaces
- Deleted 16 legacy blueprint JSON fixtures + 20 corresponding interface stubs
- Deleted 7 tests whose assertions were V1/V2-specific (`AftermarketTest`, `Cip113Test`, `BasicOptionBlueprintCompileTest`, `RecurringPaymentBlueprintCompileTest`, `SundaeSwapV2Test`, `SundaeSwapTest`, `PairBlueprintCompileTest`) plus 2 obsolete (`BluePrintTest` was 100% commented out)
- Trimmed `$`-syntax test cases from `BlueprintAnnotationProcessorTest`, `FieldSpecProcessorTest`, `BlueprintUtilTest`, `JavaFileUtilTest`, `DefaultNamingStrategyTest`

**After-the-fact audit** restored two over-deletions:
- `BlueprintCompileTest` — multi-blueprint co-compile smoke, restored against kept V3 fixtures (`SundaeSwapV3`, `UVerify`, `GiftCard`, `JpgStoreSniper`, `CircularNestedList`)
- `MissingTitleTest` — 2 of 3 tests cover CIP-57 *"title is optional"*; restored with renamed fixtures

Plus synthetic V3 fixtures restored `PairBlueprintCompileTest` and `MapBlueprintCompileTest` smokes.

### Phase 2 — stdlib 3.1 verification

- Added `AikenStdlib31SharedTypesBlueprint.java` + `AikenStdlib31SharedTypesTest.java` — a hand-crafted blueprint mirroring exactly what `aiken build` against stdlib 3.1 emits for shared types. The test asserts that `Address`, `PaymentCredential`, `StakeCredential`, `OutputReference`, `VerificationKeyHash`, `ScriptHash` all resolve via the existing V3 registry with no new registry code. If Aiken stdlib output ever drifts from the registered signatures, this test will fail.

### Phase 3 — 11 devnet tests covering distinct codegen surfaces

Every test runs against a real Cardano devnet (Yaci DevKit) in CI. Each test's class-level Javadoc starts with `<b>Unique trait:</b>` and `<b>Asserts on Cardano:</b>` so a reader scanning the folder sees in two sentences what each test pins down.

| Test | Unique trait | Methods |
|---|---|---|
| `LockUnlockDevnetTest` | baseline — primitive datum, no registry involvement | 1 |
| `HelloWorldDevnetTest` | one registry-resolved shared type (`VerificationKeyHash`) as a field | 1 |
| `AddressCheckDevnetTest` | composite `cardano/address/Address` — nested sum + `Optional<StakeCredential>` | 1 |
| `MultiActionDevnetTest` | 3-constructor ADT redeemer, pins constructor tags 0/1/2 | 3 |
| `OptionRedeemerDevnetTest` | top-level `Option<T>` redeemer, both `Some` and `None` arms | 2 |
| `MapDatumDevnetTest` | CIP-57 `dataType: "map"` via `Pairs<ByteArray, Int>` | 1 |
| `LockUnlockReadBackDevnetTest` | decoder direction — fetch UTxO, decode via generated converter, assert equality | 1 |
| `ParameterizedLockDevnetTest` | compile-time params via `applyParamsToScript`, distinct hashes per param | 1 |
| `MintPolicyDevnetTest` | mint purpose + `MintValidatorExtender` + Java-enum redeemer | 1 |
| `RecursiveSumDevnetTest` | self-referencing ADT (`Cons.tail: IntList`), nested-Constr depth | 1 |
| `TupleCheckDevnetTest` | `Tuple<<A,B>>` doubled-bracket key + `dataType: list` | 1 |

**14 test methods total, all PASS on CI** (run `26636162110`).

`ParameterizedLockDevnetTest` and `MintPolicyDevnetTest` initially failed in `@BeforeAll` with a Borer CBOR-decode error. Root cause: Scalus's `ScalusScriptUtils.applyParamsToScript` consumes / produces **double-CBOR-wrapped** script hex (`Program.fromDoubleCborHex` / `Program.doubleCborHex`), but the blueprint's `COMPILED_CODE` constant is **single-wrapped** and the generated validator constructor expects single-wrapped too (it re-wraps via `PlutusBlueprintUtil.getPlutusScriptFromCompiledCode`). Fixed with a tiny `applyParamsAndUnwrap` helper in each test that wraps once before Scalus and strips one CBOR layer after. The same gotcha is now explicitly called out in the user docs so the next user doesn't hit it.

### Phase 4 — user documentation gap-fill

The Plutus annotation processor already had 6 markdown user docs (`annotation-processor/docs/01..06`). This PR fills the gaps the test work surfaced:

- **New "Parameterised Validators" section** (`03-using-validators.md`) — the `(Network, String applyParamCompiledCode)` constructor and `getApplyParamHash()` method were emitted but undocumented. Includes the Scalus single-/double-CBOR-wrap gotcha with the exact error message and the helper code verbatim.
- **New "Where to find working examples" section** (`01-getting-started.md`) — points readers at the 11 devnet tests as the canonical real-network examples, with the `gradle` command to run them.
- **Top-level `Option<T>` codegen gap** documented (`02`, `06`) — the processor only emits `OptionXxxData` when `Option<T>` is a field of an enclosing constructor; users now have explicit guidance on the hand-rolled fallback.
- **Enum-redeemer instantiation pattern** (`03`) — `Data<?> r = () -> converter.toPlutusData(MintAction.Mint)`.
- **Decoder direction** (`02`) — fetch UTxO inline-datum, deserialise via generated converter, why this catches encoder/decoder drift.
- **Recursive ADT codegen** (`02`) — works without intervention.
- **Sum-type variant instantiation pattern** (`02`) — `new GreetData()` vs `new WithdrawData()`.
- **Pairs / Map clarification** (`02`) — both emit `dataType: "map"` and map to `java.util.Map<K, V>`.
- **CIP-57 doubled-bracket Tuple convention** (`02`, `06`) — `Tuple<<A,B>>`.
- **Mint workflow tightened** (`03`) — simple `mintToAddress(redeemer, asset, addr)` is now the primary example instead of the bulk-`MintAsset` form.
- **Stale `@AikenStdlib(V3)` example** collapsed to a single honest "annotation is a no-op, do not add it" paragraph.

219 insertions, 36 deletions across the 6 docs. Markdown only.

## What's NOT in this PR

- **Top-level `Option<T>` codegen.** The gap is documented; closing it in the processor is a follow-up. `OptionRedeemerDevnetTest` shows what users do today via a hand-rolled wrapper.
- **CHANGELOG / migration guide.** This PR is a real breaking change for stdlib v1/v2 users; a release-time doc covering "you must re-compile with Aiken ≥ 1.1.x" is a separate concern.
- **ADRs** in `annotation-processor/adr/blueprint/` left in place as historical decision records. Marking the V1/V2-era ones \"Status: Superseded\" is a candidate follow-up.
- **`BlueprintAnnotationProcessorTest.blueprintProcessorShouldGenerateExpectedValidatorArtifacts`** — was a snapshot test against a deleted V1 blueprint + deleted snapshot file. Recreating it against a V3 blueprint is doable but extra scope.

## Related — CIP-57 upstream amendment

The user-facing surface of \"how Aiken names parametric types in plutus.json\" was silent in CIP-57. While doing this work it became clear that other producers (Scalus, OpShin, plu-ts) are free to invent their own conventions if we don't standardise. Filed [cardano-foundation/CIPs#1203](https://github.com/cardano-foundation/CIPs/pull/1203) as a draft amendment specifying the angle-bracket grammar (`Option<Int>`, `Tuple<<A,B>>`, `Pair<K,V>`) plus JSON Pointer `~1` escaping for `$ref`. It's draft — awaiting review from upstream implementor maintainers.

## Test plan

- [x] `./gradlew :plutus-aiken:test` — green
- [x] `./gradlew :annotation-processor:test` — green
- [x] `./gradlew :annotation-processor:compileCodegenTestJava` — green
- [x] `./gradlew :annotation-processor:codegenTestNoRegistry` — green (registry-disabled mode still works)
- [x] `./gradlew :plutus-aiken:build :annotation-processor:build` — green
- [x] **Integration Tests (Devnet)** — green on CI run `26636162110`. **141 tests passed, 0 failed, 0 skipped.** All 11 plutus devnet tests with 14 methods total each individually PASS — the per-class XML reports are in the run artifacts.
- [x] Codegen Tests, Test deployment, CodeQL — all green on the latest commit


satran004 (approved):
Thanks @matiwinnetou . LGTM

PR #647, opened 2026-08-08
Add Nexus backend provider module

edridudi:
## Summary

Adds a new `backend-modules/nexus` module that implements `BackendService` by wrapping the [nexus-java-client](https://github.com/Gero-Labs/nexus-java-client) SDK — a multi-chain Cardano data provider — alongside the existing koios / blockfrost / ogmios backends. Structure mirrors the koios module.

## What's included

- `NexusBackendService` + 11 sub-service adapters (`Nexus{Account,Address,Asset,Block,Epoch,Metadata,Network,Pool,Script,Transaction,Utxo}Service`), a `Constants` class, and a shared `NexusResultMapper` that converts the SDK's `Result`/`ApiException` to the bloxbean equivalents and maps SDK models to bloxbean models.
- **Coverage highlights:** UTxOs synthesized from the address service (koios pattern); **full `MetadataService` parity** (all five methods, incl. by-label + CBOR); transaction redeemers from plutus contracts; account/asset/address history with in-memory pagination for the `(count, page)` variants.
- **Documented gaps:** data the SDK does not expose (e.g. block-by-number, epoch-by-number, account stake-history, policy assets, address rollup details) throws `UnsupportedOperationException`, consistent with how the koios module handles its own gaps.

## Dependency

`io.github.gero-labs:nexus-java-client:1.1.0` — published on **Maven Central**, Java 17, added to the version catalog as `libs.nexus.java`.

## Testing

Unit tests mock the SDK sub-services and assert model mapping, `Result` conversion, exception translation, pagination, and the unsupported-method paths. The module builds green (`./gradlew :backend-modules:nexus:build`).

Note: unlike the koios module's live-preprod integration tests, this module uses mocked-SDK unit tests (no live Nexus endpoint is assumed in CI). Happy to add a live IT harness if preferred.

Feedback welcome on conventions, coverage expectations, and whether you'd like the integration-test style aligned with the other backend modules.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #648, opened 2026-08-16
docs(nexus): add backend module README

edridudi:
## Summary

Follow-up to #647 (per @satran004's request). Adds `backend-modules/nexus/README.md` so the Nexus backend module has docs like the other backend modules.

Mirrors the koios/ogmios README structure:
- Maven / Gradle dependency coordinates (`cardano-client-backend-nexus`)
- How to construct `NexusBackendService` — `(baseUrl, apiKey, Network)` and the default-URL / no-key convenience constructor
- A short usage example
- The module build command
- Documented limitations: the endpoints that throw `UnsupportedOperationException` (verified against the merged source) and the ignored `order` argument

Docs-only change; no code touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend-modules/nexus/README.md
+++ b/backend-modules/nexus/README.md
@@ -1,0 +1,90 @@
+# cardano-client-backend-nexus
+
+### Nexus Backend implementation for [Cardano Client Lib](https://github.com/bloxbean/cardano-client-lib)
+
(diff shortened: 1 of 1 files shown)


satran004 (approved):
Thanks @edridudi

PR #649, opened 2026-08-22
feat: improve TxStream API developer experience

satran004:
## Summary

This PR improves the TxStream developer experience through progressive disclosure while preserving the existing correctness-first runtime and advanced configuration paths.

The common path now needs only a backend, an account or wallet reference, a stable item ID, and a `TxPlan`. Advanced applications can continue to construct and configure `FlowEngine` and `TxFlowStream` directly.

### Simpler setup and lifecycle

- Add the optional `FlowRuntime` composition root, which owns beginner executors, signer registration, one standard `FlowEngine`, opened streams, and deterministic close ordering.
- Add `FlowEngine.builder(BackendService)` plus account/wallet registration helpers.
- Add exception-safe `TxFlowStream.open(...)` and `Builder.open()` factories that start the stream and abort partial startup on failure.
- Let streams inherit the engine's caller-owned execution executor while preserving explicit executor overrides.

### Safe defaults and common submission API

- Default streams to `LanePolicy.byFundingSource()`, deriving FIFO lanes from each plan's syntactic funding source.
- Add `submit(String itemId, TxPlan)` and `trySubmit(String itemId, TxPlan)`; the stable item ID is also the idempotency key.
- Keep `TxWorkItem`, explicit lanes, custom planners, sources, durability, ownership, templates, and executor configuration available for advanced use cases.

### Honest outcomes and recovery

- Add `awaitSettled(...)` and `awaitConfirmed(...)` receipt helpers.
- Add typed failed, cancelled, uncertain, and timeout outcomes without treating `RECOVERY_REQUIRED` as success.
- Add explicit `awaitResolution(...)` polling for uncertain items; it reconciles known work and never resubmits it.
- Report eager validation and registration failures as rejection (`REJECTED`/typed exception) without publishing a receipt or incrementing accepted/failed counters.
- Add `onItemRejected(...)` and `onStreamAborted(...)` listener contracts.

### Throughput, diagnostics, and documentation

- Add planner-local `perWindow(ChainingMode)` support for explicit sequential or pipelined execution.
- Add the public `TxStreamCodes` catalog for stable core error-code matching.
- Add a progressive getting-started guide, effective-default documentation, updated package/module docs, and ADR 0005 describing the implemented API decisions.
- Add ADR 0006 to define the separate, proposed durable-registration and hydration work; this PR does not claim those unimplemented durability guarantees.
- Add Java 21 unit and DevKit integration lanes while retaining Java 17 compatibility.

## Preview qualification fixes

- Snapshot accepted TxPlan/FlowStep payloads so caller mutation cannot change queued transaction content or disagree with the accepted fingerprint.
- Let awaitResolution classify stored terminal results and retain known uncertainty instead of reporting a stored item unknown.
- Expose FlowRuntime.maxInMemoryIdempotencyClaims and document the default 10,000-claim runtime-wide lifetime limit.
- Enable a bounded between-execution backend output-visibility check for backend-based engines. Delayed indexing holds the lane; timeout fails only the new work before engine start.
- Qualify public durability claims and document provider consistency, indexing, retention, and incomplete-registration limits. Deeper durable registration/hydration changes are isolated in stacked follow-up #659.
- Add deterministic regressions and a DevKit test that queues four same-lane payments through the managed runtime without the soak application's submission gate.

Local validation: TxFlow unit tests and the managed-runtime DevKit integration tests pass on Java 17 and Java 21; RDBMS unit tests also pass on Java 21. See txflow/docs/TXSTREAM_PREVIEW_READINESS.md for the preview scope and qualification limits.

## Compatibility

- Add public `FlowEngine.maintenanceExecutor()`, a read-only accessor for the caller-owned executor; exposing or inheriting it does not transfer lifecycle ownership.

Most changes are additive or defaulted. Existing explicit `.lane(...).executor(...).build(); start();` usage remains supported, and explicit builder options take precedence over defaults.

Intentional preview behavior changes:

- Omitting a lane now derives one from the funding source.
- Omitting a stream executor inherits the engine executor when available.
- Eager invalid work is rejected instead of returning `OK` with an already-failed receipt.
- Blocking submission to a standby stream reports `TXSTREAM_NOT_ACTIVE`.
- `byFundingSource()` replaces `byFundingAddress()` as the preferred name; the old method remains as a deprecated forwarding alias for one preview release.

## Validation

- Expanded unit and contract coverage for runtime ownership, executor inheritance, lane defaults, startup cleanup, receipt waits, rejection semantics, abort notifications, pipelining, and error-code catalog consistency.
- Added Yaci DevKit integration coverage for the documented beginner path and TxStream behavior.
- CI verifies TxFlow on Java 17 and Java 21, including the managed-runtime beginner flow on DevKit.
- CI covers Clean Build, CodeQL, SonarCloud, DevKit integration tests, TxFlow RDBMS qualification, and Java 21 compatibility; consult the checks on the latest head for current results.

## Visibility and lifecycle guarantees

Visibility retries release the dispatch worker and in-flight slot between probes while preserving lane FIFO. Parked work retains its buffer permits, so `maxBufferSize` includes every waiting item. The scheduler only wakes the lane; backend I/O runs through the worker executor. FAILED/CANCELLED repair clears the affected lane's item hold directly, and stale timers cannot redispatch it.

Plain `close()` preserves retry timers until drain completes, then cancels remaining timers. If the wait is interrupted, it aborts queued work, signals running executions and preserves TXSTREAM_INTERRUPTED and the thread interrupt flag; timer cleanup still runs. Abort cancels them immediately and consistently reports TXSTREAM_ABORTED, including when observed after a probe. This prevents a parked lane from losing its only wakeup during graceful close.

FlowRuntime supplies the scheduler. Direct stream builders inherit the engine's maintenance executor when it implements ScheduledExecutorService; an explicit stream scheduler takes precedence. If neither provides a scheduler, an unsuccessful probe fails before engine start. A plain Executor cannot provide delayed retries. Neither inherited executor is shut down by the stream.

Backend-based engines used through TxStream enable the gate by default; custom historical-output services can explicitly disable backendVisibilityChecks. Ordinary FlowEngine execution does not invoke the gate. A timeout fails every member of the waiting execution before engine start, retaining registrations; new-ID retry applies only to that known pre-start failure. Pending hashes are process-local and reset on restart.

Java 17 and Java 21 TxFlow unit suites passed, including the reproduced plain-close regression, inherited/overridden scheduler selection, buffer-capacity retention, unrelated-lane progress, repair wakeup, abort status and timer cleanup.


txflow/src/main/java/com/bloxbean/cardano/client/txflow/stream/EngineTxFlowStream.java
```diff
@@ -1236,21 +1244,32 @@ public TxStreamReceipt submit(TxWorkItem item) {
...
+                notifyRejected(item, acceptance.rejection);
+                throw acceptance.rejection;
             case CONFLICT:
                 throw acceptance.conflict;
             case PAUSED:
-                // Blocking submit has no park-and-retry contract; a standby
-                // refusal keeps the established TXSTREAM_CLOSED code (only the
-                // non-blocking trySubmit distinguishes PAUSED for adapters).
-                throw new TxStreamException("TXSTREAM_CLOSED",
+                throw new TxStreamException("TXSTREAM_NOT_ACTIVE",
```
satran004, line 1276: **[suggestion]** Blocking submit on a STANDBY owner now throws `TXSTREAM_NOT_ACTIVE` (tests were updated to match), but the public `Builder.ownership(...)` javadoc still tells callers that `submit` refuses with `TXSTREAM_CLOSED` (`TxFlowStream.java:931`). That is the HA front-door contract for this preview behavior change; source adapters and operators comparing `getCode()` will follow the wrong constant.

**Suggestion:** Update the ownership javadoc (and any other submit-path docs that still say standby is `TXSTREAM_CLOSED`) to `TXSTREAM_NOT_ACTIVE`, keeping `trySubmit` as `PAUSED`.

txflow/src/main/java/com/bloxbean/cardano/client/txflow/stream/TxFlowStream.java
```diff
@@ -217,6 +224,18 @@ public interface TxFlowStream extends AutoCloseable {
...
 
+    /**
+     * Submits a common single-transaction plan using the item id as its
+     * idempotency key.
+     *
+     * @param itemId stable caller-visible item and idempotency identity
+     * @param plan portable transaction plan
+     * @return receipt for the accepted item
+     */
+    default TxStreamReceipt submit(String itemId, TxPlan plan) {
```
satran004, line 255: **[suggestion]** The new beginner overload `submit(String, TxPlan)` is the documented front door, but its javadoc only describes the success return. It does not inherit the sibling `submit(TxWorkItem)` `@throws` and omits eager-validation rejection, `TXSTREAM_NOT_ACTIVE`, unstarted/`TXSTREAM_CLOSED`, and interruption. Callers who only wrap `awaitConfirmed(...)` will miss submit-time failures.

**Suggestion:** Document the same outcome table as Decision 8 (throws the typed cause; no receipt) plus standby/`TXSTREAM_NOT_ACTIVE` and not-accepting/`TXSTREAM_CLOSED`.

satran004: Addressed in 397b00e1. Expanded submit(String, TxPlan) Javadoc to cover conflicts, eager rejection with no receipt, standby/not-accepting codes, interruption, and argument errors. I also documented the parallel trySubmit(String, TxPlan) result contract.

txflow/src/main/java/com/bloxbean/cardano/client/txflow/stream/EngineTxFlowStream.java
satran004: Addressed in 397b00e1. Updated the ownership Javadoc and all stale standby references in the durability, design, DX, and readiness docs to use TXSTREAM_NOT_ACTIVE for blocking submit while retaining PAUSED for trySubmit.

Issue #650, opened 2026-08-26
QuickTx: infer certificate/vote witnesses from the built body for fee estimation (additionalSignersCount is easy to get wrong)


matiwinnetou:
@satran004 — surfaced while fixing a fee bug in cardano-client-bindings (bloxbean/cardano-client-bindings#75); flagging in case it's worth solving upstream. If you don't think so, feel free to close.

## Context

For consumers that **build unsigned** (no `withSigner`), fee estimation depends entirely on the caller passing an accurate `additionalSignersCount`. `BalanceTxBuilders.balanceTxWithAdditionalSigners` budgets:

```java
UtxoUtil.getNoOfRequiredSigners(context.getAllUtxos()) + additionalSigners
```

The UTXO half is derived automatically, but the certificate/vote half is a caller-supplied guess — and by fee-calculation time the transaction body already **contains** the information: stake/pool certificates, DRep certificates, votes, and withdrawals each imply a distinct vkey witness that will be attached later.

In the bindings we passed a role-blind guess (`max(1, txCount)`) and underpaid by exactly one witness (~4,400 lovelace at `min_fee_a=44`) for any tx combining certificate roles, e.g. `stake_registration` + `drep_registration` in one TxPlan tx → `FeeTooSmallUTxO` at submission. Measured offline (v0.8.0-pre4):

```
stake_registration only:        fee=171441  min_fee(signed)=171441  → exact
stake_reg + drep_registration:  fee=173201  min_fee(signed)=177645  → short by 4,444
```

We've fixed it on our side by inferring roles from the TxPlan intents before calling `additionalSignersCount`, so there is **no urgency** — this is a question of where the logic belongs long-term.

## Suggestion

Have the fee-calculation path (or `balanceTx`) count the witnesses implied by the built body itself — distinct stake credentials across certificates/withdrawals/vote delegations, DRep credentials, voters, pool keys — and treat `additionalSignersCount` as a supplement for what the body cannot show (e.g. native-script `sig` policies, external co-signers). That would make every build-unsigned consumer (bindings, any FFI/service user of `build()`) correct by default, and `withSigner` users are unaffected since their signers are already counted.

Happy to contribute the change if you think it belongs in CCL; equally fine keeping the inference consumer-side — in that case just close this.

matiwinnetou:
[...]

PR #652, opened 2026-08-27
spike: count body-implied witnesses in fee estimation (#650)

matiwinnetou:
## What this is

A **prototype/spike** for #650, to judge whether the juice is worth the squeeze — not a polished change. Close it freely if the answer is no.

## Problem

`balanceTxWithAdditionalSigners` budgets fee witnesses as `UtxoUtil.getNoOfRequiredSigners(inputs) + additionalSigners`. The UTXO half is automatic; the rest is a caller-supplied guess. Callers that **build unsigned** (FFI bindings, services — no `withSigner`) routinely guess wrong, and underestimation means the node rejects with `FeeTooSmallUTxO`. Measured in cardano-client-bindings (bloxbean/cardano-client-bindings#75): a tx combining a stake and a DRep certificate was short by exactly one witness (4,444 lovelace at `min_fee_a=44`); a native-script spend whose script UTXO covered the whole tx was short the script's sig witness.

By fee-calculation time the transaction already shows most witnesses it will need. This spike makes the library count them.

## Change

- **`WitnessCountEstimator.countBodyImpliedSigners(Transaction)`** — distinct key-hash credentials across: certificates (pre-Conway stake certs + Conway `RegCert`/`UnregCert`/vote-deleg combos, DRep certs, committee certs, pool registration/retirement), withdrawals, `votingProcedures` voters, `requiredSigners`, and native-script `ScriptPubkey` trees. Script-hash credentials are skipped (script-witnessed, not vkey). `any`/`atLeast` scripts deliberately over-count — overpaying is safe, underpaying is a rejection.
- **`balanceTxWithAdditionalSigners`** now uses `utxoSigners + max(additionalSigners, bodyImplied)` — `max` (not sum) so existing callers passing an accurate count aren't double-budgeted, and `additionalSigners` stays as the override for what the body cannot show (e.g. scripts via reference inputs).

## Open questions for review

1. **`max` vs sum vs replace** — `max` is the backward-compatible middle ground; happy to change.
2. **Pool owners** — `PoolRegistration.poolOwners` also must witness; noted in the code but not counted in the spike.
3. **Legacy certs** (genesis, MIR) skipped.
4. Whether this belongs in `FeeCalculators` instead so every fee path benefits, not just `balanceTx*`.
5. `withSigner` users are unaffected in correctness but could see slightly higher (safer) fees where their signer count was lower than what the body implies.

## Tests

New `WitnessCountEstimatorTest` (cert/required-signer/native-script counting, dedup, script-hash skipping). `:function:test` and `:quicktx:test` pass unchanged.

### DIFF
--- a/function/src/main/java/com/bloxbean/cardano/client/function/helper/BalanceTxBuilders.java
+++ b/function/src/main/java/com/bloxbean/cardano/client/function/helper/BalanceTxBuilders.java
@@ -107,10 +107,20 @@ public class BalanceTxBuilders {
         return (context, txn) -> {
             String changeAddress = changeAddressIter.getFirst().getAddress();
 
-            FeeCalculators.feeCalculator(changeAddress, UtxoUtil.getNoOfRequiredSigners(context.getAllUtxos()) + additionalSigners).apply(context, txn);
+            // The transaction itself already shows most witnesses it will need (certificates,
+            // withdrawals, votes, required signers, native scripts) — count them, so callers that
+            // build unsigned and cannot attach signers get a correct fee by default (#650).
+            // max() rather than sum keeps existing callers who pass the count themselves from
+            // double-budgeting; additionalSigners remains the override for what the body cannot
+            // show (e.g. scripts supplied via reference inputs).
+            int bodyImpliedSigners = WitnessCountEstimator.countBodyImpliedSigners(txn);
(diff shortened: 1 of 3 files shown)


matiwinnetou:
@satran004 , I believe you have more elegant solution for this, closing.

PR #657, opened 2026-08-30
feat: add programmable-token QuickTx extension and TxPlan support

satran004:
## Summary

Implements the programmable-token architecture proposed by the CIP-113 ADR while keeping QuickTx protocol-neutral.

- adds the generic `QuickTxExtension` / build-local `TxBuildExtension` lifecycle
- adds TxPlan extension metadata, namespaces, runtime registration, schema/protocol validation, and typed intent codecs
- keeps optional `contract_version` as preserved informational provenance; it does not select behavior or reject an otherwise valid deployment
- supports `TxPlanCodec.fromYaml(yaml, runtimeVariables)` with runtime-over-document precedence
- adds `ProgrammableTokenService.txPlanCodec()` and a custom-namespace overload
- adds the top-level `programmable-token` module with a protocol-neutral API and CIP-113 adapter
- models transfer, mint, burn, register, registry update, third-party transfer, and unfrack as typed semantic intents
- preserves transaction-wide CIP-113 aggregation, deferred named-policy resolution, input reservation, index stabilization, and final verification
- adds `PlutusDataValue` for Java values, structured YAML, or CBOR hex with intent-level variable resolution
- supports role-specific YAML pairs such as `transfer_redeemer` / `transfer_redeemer_hex` and `inline_datum` / `inline_datum_hex`
- canonicalizes resolved Plutus data without requiring preservation of the original hex-vs-structured source form
- treats equivalent definite- and indefinite-length CBOR containers as the same Plutus value during redeemer aggregation
- updates ADR revision 6 with the executable TxPlan lifecycle and isolated JuLC fixture design

QuickTx has no dependency on programmable-token or CIP-113 classes. Extension intent codecs remain owned by the extension module.

## YAML example

```yaml
extensions:
  pt:
    extension: programmable-token
    protocol: cip-113
    schema_version: 1
    contract_version: 0.5.0-alpha.2 # optional informational provenance

transaction:
  - tx:
      intents:
        - type: pt:transfer
          receiver: addr_test1...
          unit: policy-and-asset-unit
          quantity: 1
          transfer_redeemer_hex: ${transfer_redeemer_cbor}
```

Structured Plutus-data templates, including nested variables, are also supported.

## TxPlan lifecycle E2E

Versioned YAML resources now cover deployment prerequisites, register + named-policy initial mint, additional mint, transfer, and full burn. The execution environment supplies dynamic values through the codec's runtime-variable map; document variables remain normal defaults. Every restored plan retains the merged mapping and contains ordinary `Tx` instances with typed programmable-token intents. No test-only placeholder syntax is used.

The lifecycle uses a custom JuLC Plutus V3 substandard rather than only an always-true validator. It validates the typed `Authorization(1, 42)` redeemer and requires the typed `HolderDatum(1, 113)` in participating programmable-token state.

The JuLC source project is checked in under `programmable-token/src/it/julc` and is intentionally isolated:

- JuLC regeneration uses Java 25
- the generated `plutus.json` is checked into integration-test resources and hash-verified
- CCL builds and runs tests on Java 17 without JuLC or Java 25 bytecode on its classpath
- JuLC dependencies exclude their transitive released CCL artifact

## Verification

- Java 25: `julc build -v .` — custom Plutus V3 validator compiled, pinned hash `09626cef...`
- isolated JuLC `runtimeClasspath` — no `cardano-client-lib` dependency
- Java 17: `./gradlew :quicktx:test :programmable-token:test` — passed
- Java 17: `./gradlew :quicktx:build :programmable-token:build` — passed
- Java 17 + Yaci DevKit: existing `Cip113EndToEndIT` and new `Cip113TxPlanResourceEndToEndIT` — passed together (18 tests total)
- direct codec tests parse unresolved templates without whole-document substitution
- structured and CBOR-hex redeemers are exercised in the same lifecycle, including semantic equality at a shared withdrawal credential

## Follow-up

- #658 tracks migration of existing core QuickTx intents to `PlutusDataValue` in a separate compatibility-focused PR.

This PR targets `feat/basic-cip-113-integration` so it can be used to discuss and refine the replacement API before PR #653 is finalized.


satran004:
## Programmable Token readiness follow-up

The full findings, decisions, rationale, and acceptance criteria are captured in [ADR-CIP113-002](https://github.com/bloxbean/cardano-client-lib/blob/cip113_review_planning/adr/cip113/002-programmable-token-readiness-and-remediation-plan.md).

We will use this checklist to address the work incrementally while preserving the existing extension architecture.

### Phase 0 — correctness guards

- [ ] Restrict a CIP-113 third-party transaction to one holder and one distinct policy; retain aggregation of multiple outputs for that policy.
- [ ] Aggregate and compare both transfer and issuance redeemers for burns; reject conflicting authorization instead of using last-write-wins.
- [ ] Reject unsupported verification-key operational credentials before registration/update can create an unusable token.
- [ ] Remove `unfrack` / `pt:unfrack` from the schema-v1 public capability surface until implemented.
- [ ] Add focused regression tests for all Phase 0 guards.

### Phase 1 — reusable build-local intent overlay

- [ ] Add build-local prepared-intent storage associated with each source transaction to `ExtensionBuildContext`.
- [ ] Add an internal overlay-aware `AbstractTx.complete(...)` path that treats authored and prepared intents as one build-time set.
- [ ] Include prepared intents in output/input calculation, intent application, deposits, script detection, and mint detection where applicable.
- [ ] Change `Cip113BuildExtension` to contribute prepared intents without mutating the authored `Tx`.
- [ ] Verify a successful or failed build does not change authored intent count or serialized TxPlan.
- [ ] Verify the same plan can be built sequentially without duplicate inputs, outputs, mints, withdrawals, redeemers, or witnesses.
- [ ] Add composition and normal-QuickTx regression tests.

### Phase 2 — service and deployment hardening

- [ ] Publish deployment resolution as one atomic immutable resolved state.
- [ ] Make concurrent first use wait for the same complete success/failure result; define retry behavior explicitly.
- [ ] Make registry/script caches safe for independent concurrent builds and prefer per-build registry snapshots.
- [ ] Remove silent fallback to the bootstrap coordination output on backend lookup failure.
- [ ] Permit bootstrap fallback only after positively verifying the output remains unspent and carries the expected NFT.
- [ ] Validate TxPlan `network` and `bootstrap_tx` against the configured deployment before chain I/O.
- [ ] Add concurrency, backend-failure, and deployment-mismatch tests.

### Phase 3 — TxPlan and semantic-model hardening

- [ ] Replace whole-document YAML string substitution with structural `JsonNode` runtime-variable resolution.
- [ ] Preserve native values for exact `${name}` nodes and safe string interpolation for embedded placeholders.
- [ ] Ensure replacement values are not recursively interpreted as templates.
- [ ] Preserve `PlutusDataValue` structured/HEX resolution, including validation after resolving variables inside `_hex`.
- [ ] Make `ExtensionMetadata` and nested deployment values immutable/defensively copied at all boundaries.
- [ ] Validate programmable-token asset names: no prefix, even valid HEX, maximum 32 decoded bytes, canonical lowercase output.
- [ ] Define extension ordering as numeric `order()` followed by stable extension id and update ADR-CIP113-001 wording.
- [ ] Add the CIP-113 compatibility manifest with source commit, blueprint hash, validator hashes, schema version, and capabilities.
- [ ] Add adversarial YAML-variable, metadata-mutation, asset-validation, and deterministic-round-trip tests.

### Phase 4 — internal simplification

- [ ] Extract an aggregate `Cip113TransactionPlanner` producing an immutable transaction-wide plan.
- [ ] Extract a `Cip113CoreIntentEmitter` for build-local ordinary QuickTx intents.
- [ ] Extract a `Cip113IndexFinalizer` for ledger ordering, embedded indexes, stabilization, and final verification.
- [ ] Remove unused copied fluent-routing overrides, eager/declaration machinery, duplicate assignments, and stale Javadocs.
- [ ] Confirm before/after transaction body and witness fixtures remain equivalent for supported scenarios.

### Integration-test hardening

- [ ] Add a transfer to a genuinely different owner's smart wallet with before/after balance assertions.
- [ ] Exercise multiple transfer intents, custom datum, and variable-bearing structured/HEX redeemers in the cross-owner TxPlan flow.
- [ ] Make DevKit reset/top-up fail on non-2xx responses.
- [ ] Use assumptions only for unavailable DevKit; fail assertions for protocol/setup/build/submission/state errors after availability is established.
- [ ] Use SLF4J with a compatible integration-test runtime provider instead of relying on `System.out` diagnostics.

### Release gate

- [ ] `./gradlew :quicktx:test :programmable-token:test`
- [ ] `./gradlew :quicktx:build :programmable-token:build`
- [ ] `./gradlew :programmable-token:integrationTest` with no protocol-related skips.
- [ ] `./gradlew clean build`
- [ ] Confirm public operations exactly match advertised capabilities and limitations are documented.
- [ ] Confirm Java 17 remains the CCL baseline and no QuickTx-to-programmable-token dependency was introduced.

programmable-token/src/it/java/com/bloxbean/cardano/client/programmabletoken/cip113/tx/Cip113EndToEndIT.java
```diff
@@ -60,15 +73,16 @@
...
  */
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+@ResourceLock("yaci-devkit")
 public class Cip113EndToEndIT {
 
     private static BackendService backendService;
     private static Account account;
     private static Address ownerAddress;
-    private static ProgrammableTokenService tokenService;
+    private static Cip113ProtocolService tokenService;
```
Kammerlo: I would rename this variable for better readability to protocolService or similar.

programmable-token/src/main/java/com/bloxbean/cardano/client/programmabletoken/intent/ProgrammableMintIntent.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Map;
+
+/** Declarative programmable-token mint. */
+@Data
+@Builder(toBuilder = true)
+@NoArgsConstructor
+@AllArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public class ProgrammableMintIntent implements ProgrammableTokenIntent {
```
Kammerlo, line 29: In normal Tx Building you use a minus amount for burning? Why is there a split needed? Just curious?

programmable-token/src/main/java/com/bloxbean/cardano/client/programmabletoken/ProgrammableTokenTx.java
```diff
@@ -0,0 +1,145 @@
...
+        return this;
+    }
+
+    @Override
+    public ProgrammableTokenTx from(Wallet sender) {
+        super.from(sender);
+        return this;
+    }
+
+    public ProgrammableTokenTx transfer(String receiver, Amount amount, PlutusData transferRedeemer) {
```
Kammerlo, line 49: Reference inputs are added as usual to the transaction right?

Kammerlo (commented):
I think the code and the devX API is quite nice and will make the life much easier. Once we have a preview version I would love to update the CIP-113 Platform to use it and build the transaction with that extension. This will help us to find flaws and iterate through the improvements.

programmable-token/src/main/java/com/bloxbean/cardano/client/programmabletoken/intent/ProgrammableMintIntent.java
satran004: Just to keep it same as mint/burn as mentioned in CIP113 spec. But we can also change it if required.

satran004: A bit more detail on why burn stays its own verb instead of a negative `mint` quantity: on CIP-113 a burn is a different transaction shape.

- The holder's smart-wallet UTxOs are spent through the programmable-logic base script, which runs the token's **transfer** logic with the transfer redeemer, and the remainder goes back to the smart wallet.
- On top of that, the issuance policy runs the token's **minting** logic with its own redeemer.

So a burn needs inputs plus two authorizations/redeemers, while a mint needs only a receiver and the minting redeemer. `BurnAuthorization` keeps those two redeemers distinct, and `ProgrammableBurnIntent` / `ProgrammableMintIntent` stay separate in the TxPlan so the plan states which logic is invoked. Under the hood the quantity is still handed to the core `Tx` as a negative mint amount, so the ledger-level effect is the same as a normal burn. The Javadoc on `ProgrammableTokenTx#burn` now documents this as well.

programmable-token/src/main/java/com/bloxbean/cardano/client/programmabletoken/ProgrammableTokenTx.java
satran004: Yes. The extension adds the protocol's reference inputs itself at build time (params/coordination UTxO, registry node, global state, published scripts), so the verbs don't take them. Anything you add with `readFrom(...)` goes through the usual `ReferenceInputIntent` path into the same body; duplicates are dropped, and the redeemer indexes are computed over the ledger-sorted set after balancing.

For reference, the extension's materializer is itself a `Tx` and just calls `readFrom` for the protocol UTxOs, e.g. in `Cip113TransactionMaterializer`:

```java
readFrom(service.coordinationUtxo());   // params / coordination UTxO
readFrom(node.getUtxo());               // the policy's registry node
readFrom(published.get());              // a script published on chain, referenced not witnessed
```

`Cip113BuildExtension.finish` then hands those intents to the build via `context.addPreparedIntent(...)`, so they are applied by the same `ReferenceInputIntent` as the caller's own `readFrom`, without touching the authored `Tx`.

Covered by `Cip113EndToEndIT.step19_callerReferenceInputsAreAppliedAsUsual`.

